PR #481, opened 2012-05-24
Provide low verbose to li3 test for skipped tests

hans-d:
li3 test now only shows that tests have been skipped, and only with verbose you can identify them. But verbose is very verbose. With this flag, test shows the normal test output followed by the skipped tests.


davidpersson:
Sorry @hans-d for not getting through this PR earlier.  I don't want to put much resources on improving the test runner any further (it's good enough).

PR #569, opened 2012-07-10
Add virtual fields

hans-d:
Summary:

Virtual fields in using methods in the Model, that are almost like the current behaviour to implement methods for an Entity. 

Virtual fields are defined in the model like `_schema` (although that is done for SQL db's by the datasource),  eg:

``` php
protected $_properties = array(
    'extra_field_1' => 'ExtraField`, 
    // expects model to have methods getExtraField, setExtraField, issetExtraField

    'extra_field_2' => array('get' => 'DoSomethingFancy', 'set' => false) 
    // expects model to have methods DoSomethingFancy and issetextra_field_2
    // set is done like for normal fields: data is added to _updated
)
```

that would save a lot of checks and makes the behaviour more explicit.

Outstanding:
- set behaviour in data\entity\Document - need some help with that
- squash commits (when everything is ready)

### DIFF
--- a/data/Entity.php
+++ b/data/Entity.php
@@ -75,6 +75,16 @@ class Entity extends \lithium\core\Object {
 	protected $_updated = array();
 
 	/**
+	 * Contains the callables for recognized virtual fields. These values will not be persisted.
+	 * Virtual fields are derived fields that are made available via functions in the model.
+	 *
+	 * @see lithium\data\Entity::_initVirtual();
+	 * @var mixed - array when holding data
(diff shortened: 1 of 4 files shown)


davidpersson:
If the getsyntax RFC made it through (doesn't seem unlikely it's put up for vote one time again), we had this natively implemented: https://wiki.php.net/rfc/propertygetsetsyntax-v1.2

d1rk:
Is the RFC accepted? I can not find any update on this, but the proposed syntax looked nice.

PR #989, opened 2013-07-19
Adding filter inheritance

joebeeson:


davidpersson:
Thanks for the PR. But I have to close this as it would make filters run twice in certain cases. This would not be controllable.

```
class A {
  function foo() {
     Filters::run(__CLASS__, ...);
  }
}

class B extends A {
  function foo() {
    parent::foo();
    Filters::run(__CLASS__, ...);
  }
}
```

PR #1002, opened 2013-08-17
Feature/action filters

blainesch:
Quite often you have a bit of logic duplicated between various controller actions. It's not ideal to make a real lithium filter since then the logic itself would exist in another file, or the logic to call a new method would exist in another file.

It was possible before to do before filters generically by making a filter on `Dispatcher::_callable`, however an after filter was not possible. You could potentially play with the entire response object using `Dispatcher::_call` but you could not change things that were `$controller->set()` from earlier.

### DIFF
--- a/action/Controller.php
+++ b/action/Controller.php
@@ -126,6 +126,43 @@ class Controller extends \lithium\core\Object {
 	 */
 	protected $_autoConfig = array('render' => 'merge', 'classes' => 'merge');
 
+	/**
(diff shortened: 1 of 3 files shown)


davidpersson:
@blainesch Thanks for your original proposal. As @nateabele put it, the feature is a too big departure from the current model. Since @davidgolding was able to fold that into a plugin (thanks for that!) I'm closing this PR.

PR #1070, opened 2014-02-22
Advanced cache key generation

davidpersson:
Repurposing `Cache::key()`. It now deals with advanced cache key generation
and also makes keys safe to use with given adapter configuration.

BC break: Method signature of `Cache::key()` changed. This method
was probably used seldom because of its very limited featureset.

BC break of undocumented feature: key generator functions
cannot be passed to `Cache::{read,write,delete,increment,decrement}()`
anymore. See explanation below.

- Method signature of `Cache::key()` changed. The first
  parameter is now - as in other cache methods - the name
  of the cache configuration to use.

  ```
  Cache::key('post', $data); // old
  Cache::key('default', 'post', $data); // new
  ```

- Decoupling cache key generation.

  Advanced key generation has been moved out of
  `Cache::{read,write,delete,increment,decrement}()`. Key generator
  functions may not be passed to these methods anymore. Instead generate
  the key beforehand using the new `Cache::key()`.

  Key generator functions for read and delete needed to be implemented
  in a different way then those for i.e. write as data could not be
  passed to them in any case.

  ```
  $data = array('id' => 23);

  // old: removed
  Cache::write('default', function($data) { // ... }, $data);

  // new: 2 step process
  $key = Cache::key('default', function($data) { // ... }, $data);
  Cache::write('default', $key, $data);
  ```

- The method can now be used to generate safe cache keys.

- The method can now be used to suffix cache keys with
  hashes of data.

- Adding `keys()` method for adapters.

- Adding tests.

- Adding documentation.

### DIFF
--- a/analysis/logger/adapter/Cache.php
+++ b/analysis/logger/adapter/Cache.php
@@ -81,11 +81,13 @@ class Cache extends \lithium\core\Object {
 		$config = $this->_config + $this->_classes;
 
 		return function($self, $params) use ($config) {
-			$params += array('timestamp' => strtotime('now'));
-			$key = $config['key'];
-			$key = is_callable($key) ? $key($params) : String::insert($key, $params);
-
 			$cache = $config['cache'];
+			$params += array('timestamp' => strtotime('now'));
+
+			if (!is_callable($key = $config['key'])) {
+				$key = function($data) use ($key) { return String::insert($key, $data); };
+			}
+			$key = $cache::key($config['config'], $key, $params);
 			return $cache::write($config['config'], $key, $params['message'], $config['expiry']);
 		};
 	}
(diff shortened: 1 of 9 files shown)


davidpersson:
@nateabele I'd like to merge this in 1.1, it contains some very useful features.

> Why not just pass functions passed through 'key' to be passed to key() internally?
Consistency and Complexity. First, functions passed to write will be called with $data as an param, while functions passed to read will naturally not have access to $data. Second, write now accepts an array of key/data pairs to do multi-key writes. A function as a key (generator) would be limited to single key/data writes. Functions as keys in an array are not possible.

(Rereading my comments above they partially sound too ambitious, sorry for that.)

davidpersson:
Merged in 1.2 as 8a6bf1e0e27b7b7a42e72dffb7493d2a4d49880d

PR #1110, opened 2014-08-07
Adding drain option to disable auto stream reading in action request

davidpersson:
Adding drain option to action request to allow disabling auto stream reading.

Future versions may disable draining by default.

When the drain option is disabled will not drain stream when stream data was sent.
This reduces memory usage and enables streaming very large files (i.e. uploading
video content). You must read than read stream data manually:

```
$stream = fopen('php://input', 'r');
// Do something with $stream.
fclose($stream);
```

When drain is disabled automatic content decoding is not supported anymore. It must
be manually decoded as follows.

```
// ...
$data = Media::decode($request->type, stream_get_contents($stream));
// ...
```

### DIFF
--- a/action/Request.php
+++ b/action/Request.php
@@ -163,6 +163,10 @@ class Request extends \lithium\net\http\Request {
 	 *        - `'env'` _array_: Defaults to `array()`.
 	 *        - `'globals'` _boolean_: Use global variables for populating
 	 *          the request's environment and data; defaults to `true`.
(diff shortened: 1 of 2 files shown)


davidpersson:
Merged in aadda142e7a77787ef8185b1f30a7771fb070035 and into 1.1. 
I'll decide at a later point in time, if the app behavior should already be changed to disable draining.

PR #1140, opened 2014-10-08
Improve access to methods using same configuration in adaptable classes 

farhadi:
In using Adaptable classes usually we use only one configuration at a time, but we always have to pass the first parameter to specify the desired config. So I have a suggestion to make this task less tedious using the following API (along with the current API):

```
$cache = Cache::instance('default');
$cache->read('key');
$cache->write('key', 'value');
$cache->delete('key');
```

Current API:

```
Cache::read('default', 'key');
Cache::write('default', 'key', 'value');
Cache::delete('default', 'key');
```

This pull request is a proof of concept and if you agree, I will write docs and tests for it.

Update: There was a bug in the implementation, now fixed and updated.

### DIFF
--- a/core/Adaptable.php
+++ b/core/Adaptable.php
@@ -65,6 +65,36 @@ class Adaptable extends \lithium\core\StaticObject {
 	 */
 	protected static $_adapters = null;
 
+	protected $_config;
+
+	protected $_class;
+
+	protected static $_instances = array();
+
+	public function __construct($class, $name) {
+		$this->_config = $name;
+		$this->_class = $class;
+	}
+
+	public static function instance($name = null) {
+		if (!$name) {
+			if (!$names = array_keys(static::$_configurations)) {
+				return;
+			}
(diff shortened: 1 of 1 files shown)


d1rk:
What is the current status on this? @farhadi did you expand on this?

davidpersson:
@nateabele That was related to how cache adapters return their results to the Cache class.

I've marked this RFC as it would need to affect not only Cache but all adapter based classes (Log, Session, Auth, ...). This is a structural change in how we expect those classes to work. I don't find Cache::instance() as a name very appealing as it reminds of good old singleton days in CakePHP. However there might be use of this, as we need a way to make classes "less static" (i.e. for implementing Logging PSR etc.  This would not be for the upcoming version (1.0.0), but for a future minor version.

farhadi:
@d1rk not yet.
Actually I was waiting for @nateabele and @davidpersson confirmation.

Issue #1128, opened 2014-09-30
SQL error with LIMIT, ORDER and GROUP in the same request


gmercey:
Hi,

I work with a Postgres database, and I found a problem with the model finder when there are limit, order and group instruction in the same request.

     Operation::find('all', array(
            'conditions' => array(),
            'with' => array('Partner', 'Code),
            'fields' => array('Operation.*', 'Partner.name', 'COUNT(DISTINCT "Code".id) AS total_code'),
            'group' => array('Operation.id', 'Partner.name'),
            'order' => array('date_start' => 'DESC'),
            'limit' => 5
    ));

It seems that it's caused by the first request done for the pagination. Here's the SQL error :

    for SELECT DISTINCT, ORDER BY, expressions must appear in the SELECT list

(I translate the error message from french, so it might not be the exact translation)

Thanks

davidpersson:
From my first guess, this doesn't look like it's related to using LIMIT, ORDER, GROUP together. Instead it seems the error is caused by using DISTINCT in the field list.

http://stackoverflow.com/questions/12693089/pgerror-select-distinct-order-by-expressions-must-appear-in-select-list

I'll see what I can do.

davidpersson:
I cannot reproduce the issue.

@gmercey Please provide a failing test case. A good place to add one would be at:
https://github.com/UnionOfRAD/lithium/blob/dev/tests/cases/data/source/DatabaseTest.php#L912

If you cannot provide a test case, copy and paste the SQL which causes the error from the database logs. Also the actual error message would be nice to have.

agborkowski:
btw i remember when im used distinct in simple select query i had problem too with li3 orm... insted of this i used raw sql... @todo provide eg.

koushki:
@davidpersson I have a test case for produce the issue in PR #1145 , that's clear. 

This error occur in https://github.com/UnionOfRAD/lithium/blob/dev/data/source/Database.php#L635

fixed the issue in PR #1145 but don't merge .

davidpersson:
@koushki Thanks for the heads up, #1145 will merge soon.

nateabele:
Fixed.

PR #1145, opened 2014-10-28
Fixing bug in PostgreSql when we use both limit & order in hasMany relations.

koushki:
Mentioned in #1128


davidpersson:
Will also fix #1128.

PR #1159, opened 2015-01-28
Bug/router match

koushki:
I fixed this bug that had created after merge of PR #1158 .
when we use ``Router::match`` without controller param and routers be inside scope, we get this error because controller change within params.
I changed ``dispatcher::applyRules`` like before PR #1158 but I didn't break feature (continue-dispatch-rules)  which had added in that PR.

@davidpersson @nateabele , If you think this change can be good for merge I can squash both commits together.


davidpersson:
@koushki Thanks for taking care, good to have that under test coverage now.

koushki:
I squashed both commits together.

Issue #851, opened 2013-03-06
Model finders can no longer be invoked from within an instance method.


mikegreiling:
This bug was introduced in e73ffe6 (Issue #744)

Apparently, now that Model has both a `__call()` and `__callStatic()` method, when I attempt to invoke a finder from within an instance method of the model itself using the "syntactic magic" from Model's `__callStatic()`, it will go through the `__call()` method instead.

Example:

```
class Users extends \lithium\data\Model {
    public function getNext($entity, $token = null) {
        $conditions = [...];
        return static::first(compact('conditions'));
    }
}
```

It invokes `Model::__call()` instead of `Model::__callStatic()` and I get the exception `Unhandled method call 'first'.`

If I define it this way, it'll work... but it's not ideal, and the magic method used to work just fine.

```
class Users extends \lithium\data\Model {
    public function getNext($entity, $token = null) {
        $conditions = [...];
        return static::find('first',compact('conditions'));
    }
}
```

Related stackoverflow threads:
- http://stackoverflow.com/questions/13232445/call-catches-static-method-calls
- http://stackoverflow.com/questions/8757696/how-to-force-callstatic-for-unknown-static-methods-even-if-instance-method-wit

DrRoach:
I fixed this issue in #1161 and made a pull request.

PR #1161, opened 2015-02-04
Fixed static magic method calls

DrRoach:
If a magic find is called such as first, __callStatic() will be called after __call() doesn't find any results and this will find the first magic method rather than an exception being thrown by __call() saying that the method doesn't exist.

Original issue - #851


DrRoach:
I will rewrite the tests so that they match the correct exceptions if this is wanted.

PR #1162, opened 2015-02-06
Fixing bug with non-sequential child records in hasMany relation

koushki:
Here is a bug in hasMany relation when we use relation with order params .
I don't sure about this way resolved but I couldn't find better solution for this problem.
If you have better solution, please give me your opinion.

### DIFF
--- a/data/source/Database.php
+++ b/data/source/Database.php
@@ -1335,6 +1335,49 @@ abstract class Database extends \lithium\data\Source {
 	 * @return string Formatted clause.
 	 */
 	public function order($order, $context) {
+		$model = $context->model();
+		$return = $context->return();
+		$relations = $context->relationships();
+		$direction = 'ASC';
+
+		if (is_string($order)) {
+			if (preg_match('/^(.*?)\s+((?:A|DE)SC)$/i', $order, $match)) {
+				$order = $match[1];
+				$direction = $match[2];
+			}
+			$order = array($order => $direction);
+		}
+
+		if (is_array($order) && $relations && $model && (!$return || $return === 'item')) {
+			foreach ($relations as $relation) {
+				if ($relation['type'] !== 'hasMany') {
+					continue;
+				}
+				$orders = array(
+					'main' => array(),
+					'relations' => array()
+				);
+				$keyDirection = 'ASC';
+				$relPattern = '/^(' . implode('|', array_keys($relations)) . ')\./';
+				$keyPattern = '/^(' . $model::meta('name') . '\.)?' . $model::key() . '/';
+				foreach ($order as $column => $dir) {
+					if (is_int($column)) {
+						$column = $dir;
+						$dir = $direction;
+					}
+					if (preg_match($relPattern, $column)) {
+						$orders['relations'][$column] = $dir;
+					} elseif (preg_match($keyPattern, $column)) {
+						$keyDirection = $dir;
+					} else {
+						$orders['main'][$column] = $dir;
+					}
+				}
+				$orders['main'][$model::key()] = $keyDirection;
+				$order = array_merge($orders['main'], $orders['relations']);
+				break;
+			}
+		}
 		return $this->_sort($order, $context);
 	}
 
@@ -1363,10 +1406,6 @@ abstract class Database extends \lithium\data\Source {
 		$model = $context->model();
 
 		if (is_string($field)) {
-			if (preg_match('/^(.*?)\s+((?:A|DE)SC)$/i', $field, $match)) {
-				$field = $match[1];
-				$direction = $match[2];
-			}
 			$field = array($field => $direction);
 		}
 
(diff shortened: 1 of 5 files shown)


davidpersson:
How does this compare to the following?

```
Galleries::find('all', [
    'order' => ['name' => 'DESC'],
    'with' => ['Images' => [
         'order' => ['title' => 'DESC]
    ]
]);
```

koushki:
I explain this bug with example.
Galleries table:
<table>
<tr><td>id</td><td>name</td></tr>
<tr><td>1</td><td>Foo Gallery</td></tr>
<tr><td>2</td><td>Bar Gallery</td></tr>
</table>

Images table:
<table>
<tr><td>id</td><td>gallery_id</td><td>title</td></tr>
<tr><td>1</td><td>1</td><td>Amiga 1200</td></tr>
<tr><td>2</td><td>1</td><td>Srinivasa Ramanujan</td></tr>
<tr><td>3</td><td>1</td><td>Las Vegas</td></tr>
<tr><td>4</td><td>2</td><td>Silicon Valley</td></tr>
<tr><td>5</td><td>2</td><td>Unknown</td></tr>
</table>

when we run find like below code :
```php
$galleries = Galleries::find('all', [
	'order' => ['Images.title' => 'DESC'],
	'with' => 'Images',
]);
```
Then result is :
```php
[
	2 => [
		'id' => 2,
		'name' => 'Bar Gallery',
		'images' => [
			4 => [
				'id' => 4,
				'title' => 'Silicon Valley',
			],
		],
	],
	1 => [
		'id' => 1,
		'name' => 'Foo Gallery',
		'images' => [
			3 => [
				'id' => 3,
				'title' => 'Las Vegas',
			],
			1 => [
				'id' => 1,
				'title' => 'Amiga 1200',
			],
		],
	],
]
```
But we expected :
```php
[
	1 => [
		'id' => 1,
		'name' => 'Foo Gallery',
		'images' => [
			2 => [
				'id' => 2,
				'title' => 'Srinivasa Ramanujan',
			],
			3 => [
				'id' => 3,
				'title' => 'Las Vegas',
			],
			1 => [
				'id' => 1,
				'title' => 'Amiga 1200',
			],
		],
	],
	2 => [
		'id' => 2,
		'name' => 'Bar Gallery',
		'images' => [
			5 => [
				'id' => 5,
				'title' => 'Unknown',
			],
			4 => [
				'id' => 4,
				'title' => 'Silicon Valley',
			],
		],
	],
]
```
This bug happened because ``lithium\data\collection\RecordSet::_mapRecord`` crawl next records until main table fields didn't changed and when we use order (or any other reason) that causes datas disordered, we get this bug.

I don't understand that your opinion have will fix this bug.

davidpersson:
@koushki Thanks for the great explanation and the visual examples. I now understand a lot better.

I believe there is a bug but I'm also currently wondering if the following usage should be allowed at all, and throw an exception.

```
$galleries = Galleries::find('all', [
    'order' => ['Images.title' => 'DESC'],
    'with' => 'Images',
]);
```

Cake's ORM for example doesn't support sorting on a hasMany relation and simply errors out. To me having `order => Images.title` on Galleries also doesn't make real sense.

@nateabele I'm unsure what the design goals actually were and if this just is a bug triggered by - what should be - invalid usage?

nateabele:
> Cake's ORM for example doesn't support sorting on a hasMany relation and simply errors out. To me having order => Images.title on Galleries also doesn't make real sense.

Well, it's a contrived example, I'm sure there are many valid use cases, i.e. sorting posts by top comment score, etc.

> I'm unsure what the design goals actually were and if this just is a bug triggered by - what should be - invalid usage?

The design goal was just to group child records by parent records, and the existing implementation is very naïve. I've thought of a much better & simpler one. I'll implement it this week, and use the tests in this PR to validate. Thanks @koushki!

farhadi:
This problem can also occur without having an `order` in your find.
Actually @koushki found this bug in one of our production servers in a case where there is no order at all.
And that's because without having an order, due to mysql internals, the result might be in an arbitrary order which can lead to a similar unexpected result.

koushki:
@farhadi's right, in mysql 5.6 joins may be disorder. below example is clear for my mean.

```sql
INSERT INTO `galleries` (`id`, `name`) VALUES
(1, 'Foo Gallery'),
(2, 'Bar Gallery');

INSERT INTO `images` (`id`, `gallery_id`, `image`) VALUES
(1, 1, 'someimage.png'),
(2, 2, 'image.jpg'),
(3, 1, 'photo.jpg'),
(4, 2, 'picture.jpg'),
(5, 2, 'unknown.gif');

SELECT * FROM `galleries` LEFT JOIN `images` ON `galleries`.`id` = `images`.`gallery_id`;
+----+-------------+------+------------+---------------+
| id | name        | id   | gallery_id | image         |
+----+-------------+------+------------+---------------+
|  1 | Foo Gallery |    1 |          1 | someimage.png |
|  2 | Bar Gallery |    2 |          2 | image.jpg     |
|  1 | Foo Gallery |    3 |          1 | photo.jpg     |
|  2 | Bar Gallery |    4 |          2 | picture.jpg   |
|  2 | Bar Gallery |    5 |          2 | unknown.gif   |
+----+-------------+------+------------+---------------+
```

‌@nateabele IMO child records must be order in parent records and the developer can use below code if he want to order posts by top comment score

```php
Comments::all([
	'with' => 'posts',
	'order' => ['score']
]);
```

davidpersson:
I've restarted the travis build. Some jobs were hanging, now everything is back green again.

davidpersson:
There are actually two problems here. 1. Non sequential records and 2. Ordering in/by? has many.

davidpersson:
We probably "just" have to fix RecordSet::_mapRecord(). If there's wrong assumption in there (sequential keys) then the method must be changed to remove the wrong assumption.

davidpersson:
Closing in favor of #1189. We're getting close :)

PR #1169, opened 2015-03-10
Test against PHP 7.0

alexbowers:
This test will allow people to see whether Li3 will work with php 7; which is expected in October.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
 
 matrix:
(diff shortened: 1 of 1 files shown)


davidpersson:
PHP 7.0 will be supported. However it's and its extensions are not ready for prime-time yet. 

Can you fix some of the failed builds caused by extension build configurations? I see that memcached and redis exts are not loaded correctly.

bildschirmfoto 2015-03-14 um 14 58 32

Please also exclude the xcache run from the PHP 7 matrix. I don't expect them to release a compatible version soon.

Thanks!

davidpersson:
Added in 1.1 b4e5b6f457eb72f75daa9feb2b2f90196c3557bb

alexbowers:
Do we have a feature list for 1.1?

davidpersson:
No not yet. 

Please note that this is a bug tracker, we cannot have general discussions here. Those are best conducted on Twitter or IRC.

PR #1170, opened 2015-03-11
Improve database encoding methods.

davidpersson:
Changed: encoding() may now throw exceptions and needs try/catch in
userland if invoked directly.

- Do not silence exceptions, allow exceptions to buble up.
- Check return value of encoding in connect.
- Refactored methods.
- Now all permutations of utf-8/UTF8 are allowed.
- Add more encoding tests.


davidpersson:
Mistakenly added branch to the wrong repo. Will reopen new PR.

PR #1172, opened 2015-03-13
Improve database encoding, refactor timezone and searchPath methods.

davidpersson:
Improve database encoding, timezone and searchPath methods. …
Changed: encoding(), timezone() and searchPath() may now throw exceptions
userland code needs try/catch if invoking these methods directly.

- Do not silence exceptions, allow exceptions to bubble up.
- Check return value of methods in connect.
- Refactored methods.
- Now all permutations of utf-8/UTF8 are allowed in encoding method.
- Add more encoding tests.

### DIFF
--- a/data/source/Database.php
+++ b/data/source/Database.php
@@ -126,11 +126,14 @@ abstract class Database extends \lithium\data\Source {
 	protected $_cachedNames = array();
 
 	/**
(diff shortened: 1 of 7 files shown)


davidpersson:
Merged in 173edb6db8a560a3f5ce61059c77df7468d41332 into 1.1

Issue #998, opened 2013-07-25
Form Signatures listing different fields between form render and submit, hashing not working


cinaeco:
I was trying to use form signing when I ran into issues:

1. `template\helper\Security`'s filter on '_defaults' also picks up label names e.g. `$state[$id]['fields']` might contain `'AddressesPostCode'` and `'address.postCode'` field names, whereas when checking the request data, only the `'address.postCode'` field would exist. Signatures on forms with labels would never agree. I bypassed this temporarily by putting in a check for `$params['method'] == 'label'` and skipping those elements.

2. When `FormSignature::check()` calls `FormSignature::key()`, it has it's fields array with field names as values, instead of as keys as `template\helper\Security`'s filters do. I bypassed this temporarily by putting in an `array_flip()` before calling `key()`.

3. I then tried to submit a form, but first mangled a field name, thinking this would produce an error, but it didn't. The hashes for submitting a form with and without a mangled field name remain the same e.g.

original field (1 out of 14 others): 'parent.mobile'
original hash: '$2a$10$NuNTOeXv4OHpPJtbdAmfReEtZeKhI9F9xBL8qGvbO.kFHM4KzJoRK'
 
mangled field (1 out of 14 others): 'parensfsdfsdfsdfsdft.mobile'
mangled hash: '$2a$10$NuNTOeXv4OHpPJtbdAmfReEtZeKhI9F9xBL8qGvbO.kFHM4KzJoRK'

Have I misunderstood how this is supposed to work? Should the field names make a difference to the hash?

nateabele:
Fixed.

PR #1173, opened 2015-03-14
Changing form signature calculation logic. Fixes #839, #998. BC-break.

### LINKED ISSUE #839: Reduce overhead in Form Signing
Hello guys

The current implementation of CSRF protection in Lithium seems to have a lot of unneeded overhead. It uses sha512 by default, but even if you define md5 as a type, you are still getting tokens generated by 'php::crypt'. This can add up if you use a couple of forms one a page. 

Could we not accomplish the same functionality by just using plain-old md5 tokens? 

Please do correct me if I'm wrong about this :)

Best regards,

Ciaro

davidpersson:
The calculation logic now uses HMAC which requires a secret key. Such
a unique key must be set in userland before starting to use form
signature generation or verification.

The new logic is basically modelled after message signatures and its
purpose is to ensure integrity of the compiled form signature string.

Adding test to prove overflowing is prevented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Fixed bug when using LIMIT, ORDER and hasMany relations with PostgreSql. #1145 (Hamid Reza Koushki)
 - Matched implementation with documentation of `Request::accepts()`. The method now returns a boolean 
   when type is provided. #1180, #856 (David Persson, David Rogers)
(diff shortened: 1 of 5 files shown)


nateabele:
@davidpersson Seems like Travis keeps failing. Did it pass before? Should we just merge it?

davidpersson:
One test was date dependent, fixed that and restarted some builds that were hanging. Should be all green in couple of minutes.

Issue #698, opened 2012-11-27
RFC: New MongoClient class of MongoDB Native Driver version 1.3.0


ericcholis:
There is a shift coming in the MongoDB Native Driver for PHP:
http://blog.mongodb.org/post/36666163412/introducing-mongoclient?yey

Specifically as of version 1.3.0, [Mongo](http://us2.php.net/manual/en/class.mongo.php) will be deprecated and replaced with [MongoClient](http://us2.php.net/manual/en/class.mongoclient.php)

There's a few new features, that I'm aware of:
- Write Concern, which will replace safe and fsync.
  - `array("w" => 3)`
- Read Preference which provides better control than setSlaveOkay over replica sets.
  - `array("readPreference" => MongoClient::RP_PRIMARY_PREFERRED)`

(see: http://us2.php.net/manual/en/mongoclient.construct.php for reference)

PR #1179, opened 2015-03-26
Fix for deprecated MongoClient::$connected property

tmaiaroto:
You know I thought this (https://github.com/UnionOfRAD/lithium/issues/698) was taken care of a while ago, but I think there were two issues (or now are). First was just the use of ```MongoClient``` instead of just ```Mongo``` which I see is in the dev branch. Second, the ```MongoClient::$connected``` property is now deprecated so checking it on ```connect()``` in the adapter is throwing deprecation warnings.

I discovered that ```MongoClient::getConnections()``` will solve the same problem, but return more reliable information (apparently connected boolean could be lying) and more of it. Should there not be any connections, this would return empty. See more here: http://php.net/manual/en/mongoclient.getconnections.php

So to resolve the deprecation warnings a simple switch on one line should do it.

This is a little awkward to test (and we're now testing PECL extension versions), but it seems to have worked and given what the change involved I imagine it's fine without breaking the call out to test? Without backward compatibility as well? Because ```getConnections()``` has been around since version 1.3.0 which was released in November of 2012. Though I recognize this may bump up a minimum requirement for the framework.

In the interim while this is sorted out, one could create their own Mongo adapter and extend the MongoDb class and add their own ```connect()``` method which would have nearly the same code, but with this small change. That is to say if this deprecation warning is problematic for you.

Hopefully this should clear some things up.

### DIFF
--- a/data/source/MongoDb.php
+++ b/data/source/MongoDb.php
@@ -283,7 +283,7 @@ class MongoDb extends \lithium\data\Source {
 	 * @return boolean Returns `true` the connection attempt was successful, otherwise `false`.
 	 */
 	public function connect() {
-		if ($this->server && !empty($this->server->getConnections()) && $this->connection) {
+		if ($this->server && count($this->server->getConnections()) > 0 && $this->connection) {
 			return $this->_isConnected = true;
 		}
 
--- a/data/source/MongoDb.php
+++ b/data/source/MongoDb.php
@@ -283,7 +283,7 @@ class MongoDb extends \lithium\data\Source {
 	 * @return boolean Returns `true` the connection attempt was successful, otherwise `false`.
 	 */
 	public function connect() {
-		if ($this->server && count($this->server->getConnections()) > 0 && $this->connection) {
+		if ($this->server && $this->server->getConnections() && $this->connection) {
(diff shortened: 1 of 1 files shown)


davidpersson:
`MongoDb::disconnect()` already uses `getConnections()`, so it seems to me the old property check was in `connect()` was left from a previous change. Just a tiny suggestion: do we really need count() or isn't the check if it's true'ish enough?

tmaiaroto:
It always returns an array. I used if !empty() but that's a problem for older PHP versions. If there's a more elegant or preferred way, let me know.

In theory we could be doing deeper checks too since that returns all the connections. So "is the master node connected" ? For example...because it may mean things are read only. Though that's a whole separate concern I know.

tmaiaroto:
I've updated it to work exactly like ```disconnect()``` I've been working in other languages and forgot that would work =)

davidpersson:
I know that feeling :)  For our abstraction we go far enough with connection state checking. Users can still access the connection object directly and operate on it directly if they have more needs.

tmaiaroto:
Yup, sounds good.

PR #1171, opened 2015-03-11
Allow enabling MySQL strict mode.

davidpersson:

PR #1181, opened 2015-03-31
Disable Number to String conversion when the MySQL Native Driver is used

ericcholis:
Allows the MySQL Native Driver to cast resulting numeric fields correctly.

I'm unsure if this is the best place or method to accomplish this.

### DIFF
--- a/data/source/database/adapter/MySql.php
+++ b/data/source/database/adapter/MySql.php
@@ -320,6 +320,11 @@ class MySql extends \lithium\data\source\Database {
 			$options = $params['options'];
 			$conn->setAttribute(PDO::MYSQL_ATTR_USE_BUFFERED_QUERY, $options['buffered']);
 
+			if (strpos($conn->getAttribute(PDO::ATTR_CLIENT_VERSION), 'mysqlnd') !== false) {
+				$conn->setAttribute(\PDO::ATTR_EMULATE_PREPARES, false);
+				$conn->setAttribute(\PDO::ATTR_STRINGIFY_FETCHES, false);
+			}
+
 			try {
 				$resource = $conn->query($sql);
 			} catch (PDOException $e) {


davidpersson:
I know this is probably hard to test, but we need a test with each feature addition.

ericcholis:
Not 100% sure how to approach a test like this.

My initial thought is to create a `MySqlTest::testDriverTypeCasting` integration test. Would work similarly to `MySqlTest::testDefaultValues`, but skip if `mysqlnd` isn't detected and do `assertIdentical` rather than `assertEqual`.

Any thoughts/suggestions would be greatly appreciated.

davidpersson:
1. Tests that break because of the change probably need relaxed assertions. 
2. Base your tests off https://github.com/UnionOfRAD/lithium/blob/dev/tests/integration/data/source/database/adapter/MySqlTest.php#L322
3. Create two tests one which skips when mysqlnd is used the other one when it is not used.

Generally tests should contain as little logic as possible.
Normally we stand by "Never change existing tests." unless a new change reveals that these tests are too fragile/strict/wrong.

data/source/database/adapter/MySql.php
```diff
@@ -320,6 +320,11 @@ protected function _execute($sql, array $options = array()) {
 			$options = $params['options'];
 			$conn->setAttribute(PDO::MYSQL_ATTR_USE_BUFFERED_QUERY, $options['buffered']);
 
+			if (strpos($conn->getAttribute(PDO::ATTR_CLIENT_VERSION), 'mysqlnd') !== false) {
+				$conn->setAttribute(\PDO::ATTR_EMULATE_PREPARES, false);
```
davidpersson: Why does this need to be switched off?

davidpersson:
As a question: should we also cast non-mysqlnd results for portability reasons? I have the feeling we're going too far with that?

data/source/database/adapter/MySql.php
ericcholis: According to http://blog.ulf-wendel.de/2008/pdo_mysqlnd-the-new-features-of-pdo_mysql/

> mysqlnd returns native data types when using Server-side Prepared Statements, for example an INT column is returned as an integer variable not as a string. 

My initial tests illustrated that `$conn->setAttribute(\PDO::ATTR_STRINGIFY_FETCHES, false);` worked without `ATTR_EMULATE_PREPARES`. However, everything I could find on the topic suggested including it.

ericcholis:
I've always found it odd (and annoying) that libmysql cast all results as strings. Even the default behavior of mysqlnd is to cast strings. I know li3 does a good job with ensuring types from MongoDB results. There is a bit of overhead involved in doing that for large result sets however.

Perhaps it should be optional, via the connection? Maybe this should be an RFC first...

davidpersson:
I'll relabel this issue. Also there is https://github.com/UnionOfRAD/lithium/pull/1171

PR #1184, opened 2015-05-03
Extract String RNG into Random, hashing into Hash class.

davidpersson:
1. This is to prepare the String for a rename (probably `Text`)
   these features don't fit into a `Text` class anymore.
2. These features have more to do with security.
3. The String class has too many different scopes and should
   purely deal with String/Text manipulation.
4. Packages inside the security NS have one dependency less
   outside themselves.


nateabele:
Looks good. IMO we should do it in 1.0, since semver dictates a new major version for BC-breaking changes.

davidpersson:
We don't have to as code is in place to redirect to new classes/methods. Using them just issues a deprecation warning (which can be surpressed if users want to). 

Regarding the String/Text rename. For <PHP-7.0 we can use class aliases or an empty class that extends the new one. This would not break BC for <7.0 in 1.1. 

I think it'll be hard to make 1.0 7.0 ready given the current state of affairs in php-core. It's not even decided what exactly 7.0 will be. That shouldn't hold us back from releasing 1.0. I don't want to put anymore features into 1.0, there are enough new ones already.

We can still reserve us to backport changes into 1.0 for FC.

davidpersson:
Thanks! 

Just read through the quote more thoroughly and discovered that deprecated functionality should only be removed in major versions. This doesn't affect us now. Seems everything in semver points to release more often and don't be afraid of major versions. At least that's how I read it.

nateabele:
Fair enough.

PR #1189, opened 2015-05-18
Has many fix

nateabele:
This is a work-in-progress on a fix for #1162.

More to come very shortly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,32 @@
 - Fixed bug when using LIMIT, ORDER and hasMany relations with PostgreSql. #1145 (Hamid Reza Koushki)
 - Matched implementation with documentation of `Request::accepts()`. The method now returns a boolean 
   when type is provided. #1180, #856 (David Persson, David Rogers)
-- Fixed several bugs in the FormSignature class. #839, #998, #1173 (Hamid Reza Koushki, David Persson, Ciaro Vermeire, cinaeco) 
+- Fixed several bugs in the FormSignature class. #839, #998, #1173 (Hamid Reza Koushki, David Persson, Ciaro Vermeire, cinaeco)
+- Fixed undetected buggy behavior in `Database`/`RecordSet` where records could get out of order and **associated has-many** 
+  records would be partially missing. The bug occurred when querying for results with a has-many relationship and without ordering 
+  by the primary key of the main record. In all queries involving a has-many relationship, it is adviced to first order by the main 
+  id as otherwise an exception will be thrown. 
+  We've decided against adding in the main id magically, as silently rewriting users' queries is a rabbit hole we dont't want 
+  to go down. In general implicit order by the database is not something to rely on, if you want records in a certain order, always 
+  explictly specify it. 
+  Fixes #1162 and #1182 (Hamid Reza Koushki, Nate Abele, David Persson)
+
+  ```php
+  Posts::find('all', array( // wrong :(
+	'order' => 'Comments.created',
+	'with' => 'Comments'
+  ));
+
+  Posts::find('all', array( // correct :)
+	'order' => array('id', 'Comments.created'),
+	'with' => 'Comments'
+  ));
+
+  Posts::find('all', array( // correct :)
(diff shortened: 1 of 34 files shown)


nateabele:
@davidpersson This represents the third or fourth iteration of the new interface between `RecordSet` and `Result`, and was as far as I was able to get in the time that I have. I might be able to pick it up again late next week.

davidpersson:
@nateabele This looks pretty much complete to me. I can help and add the tests from the original issue/PR (I see you have opted for throwing an exception in case of unordered key -- like it). I can also fix the array_combine issues, which probably just needs a check if the combined array is empty. Is there anything outside those points that need to be addressed?

nateabele:
> Is there anything outside those points that need to be addressed?

Nope, I just wasn't able to solve all the test fails in time (I got most of them though). I also kind of wanted to implement a way to do eager fetching, but it's not necessary (a user can always return the result directly and iterate).

davidpersson:
I've finished this branch up (refactored a bit more and fix some bugs), I've also merged @koushki's changes and tests in (they were still needed) and refactored/cleaned up those.

All in all I think the result cursors and record set classes are in a much better shape now. Way easier to understand what's going on.

@nateabele I can squash your commits together, just don't want to squash all to retain authorship information, then rebase and the branch onto dev.

nateabele:
> I've also merged @koushki's changes and tests in (they were still needed) and refactored/cleaned up those.

@davidpersson Can you explain why they were needed? I really, _really_ don't like the idea of magically rewriting peoples' queries.

davidpersson:
Then this is a misunderstanding, thought that patch was what's being extended. I also don't like  magicness. I'll remove the rewriting in `Database::order()`, and will update the new tests. Still we probably need better documentation as people will run into this issue sooner or later and if they do, there should be an explanation to solve it.

davidpersson:
@nateabele Alright the auto-magic query mangling has been removed. I'd be ready to rebase, squash and merge.

davidpersson:
Comparing runs current dev vs dev (earlies) + this patch (later)

bildschirmfoto 2015-07-12 um 16 30 10
bildschirmfoto 2015-07-12 um 16 29 40
bildschirmfoto 2015-07-12 um 16 28 13

nateabele:
> Then this is a misunderstanding, thought that patch was what's being extended. I also don't like magicness. I'll remove the rewriting in Database::order(), and will update the new tests.

Okay sounds good. Yeah I pretty much started over from scratch.

> Still we probably need better documentation as people will run into this issue sooner or later and if they do, there should be an explanation to solve it.

Agreed. I was actually thinking about having a page in the docs with a list of all exceptions thrown in core and an explanation of each. Sound good?

nateabele:
Btw, I'm not totally sure, but, the charts show that the core is _faster_ with this patch, correct?

davidpersson:
It uses less memory and less function calls. The change in CPU time isn't something I'd look at as long it isn't significant. Unfortunately xhgui has no way yet to get averages over many requests.

The function call percentage (6%) is relatively high because of repeated actions inside the requests (20 queries). All in all I think it's a good change and memory savings will be relatively higher if there's more then just one record involved per query or records have many fields/big values.

nateabele:
Awesome. Yeah, that's why I wasn't totally sure, but yeah, lots of memory usage is at least somewhat correlated to CPU, so there should be a good performance improvement on large data sets there as well.

davidpersson:
@nateabele Should I wrap this up and merge?

nateabele:
@davidpersson Yes, go for it. Looks great.

davidpersson:
Merged in 9e5351a.

koushki:
@nateabele, @davidpersson   I' m not sure that everything be correct now. As I've mentioned in PR #1162, We have incorrect result when we don't use `order` in `Mysql 5.6+`.

```php
  Posts::find('all', array( // wrong :(
	'with' => 'Comments'
  ));
```

Would you mind do something for this problem?

nateabele:
@koushki What we were suggesting is that you should always use `order` in these cases (you can set default ordering in `$_query` for simplicity).

PR #1195, opened 2015-06-14
Improve (error) exception titles in tests.

davidpersson:
This PR enables display of exception names (InvalidErrorException instead of just Exception) as well as display of severity (the name of the error level) when the exception is an ErrorException.

bildschirmfoto 2015-06-14 um 13 47 25


nateabele:
:+1:

PR #1196, opened 2015-06-16
Enable hashing of arbitrary data in Hash::calculate().

davidpersson:
Can be used to quickly generate safe cache keys:
```
Hash::calculate($entity, array('type' => 'crc32b'));
```

... or to identify closures and generate unique arrays of closures:
```
$foo = function() {};
$bar = $foo;

Hash::calculate($foo) === Hash::calculate($bar);

$map[Hash::caculate($foo)] = $foo;
$map[Hash::caculate($bar)] = $bar;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Improved
 
 - Improved database encoding, timezone and searchPath methods. #1172 (David Persson)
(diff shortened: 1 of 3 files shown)


davidpersson:
Merged in e57aabb0ebd0c906dc23c7ba6a0c6ac9e62d4393

PR #1203, opened 2015-07-08
Implementing datalist support for text fields.

davidpersson:
Text form fields now support generating corresponding `<datalist>` elements for autocompletion.

```php
$this->form->field('region', array(
    'type' => 'text',
    'list' => array('foo', 'bar', 'baz')
));
```

Also see https://www.w3.org/wiki/HTML/Elements/datalist


nateabele:
:+1:

PR #1211, opened 2015-09-05
New filter implementation (for PHP7 compat)

davidpersson:
tl;dr;

To achieve PHP7 compatibility we cannot rely on `Object` anymore. As
`Object` contained the filtering logic it needed to be extracted. The
chance has been taken to cleanly re-implement filters **in a
fully BC-compatible way**.

Features
--------
- >50 added tests, more updated and migrated tests
- full BC
- simplified modern syntax and signature
- separation of concerns filters vs chain
- dedicated namespace
- less code (excl. BC and tests)
- everything is lazy by default / chain caching
- better (future) interoperability with other middleware
  handlers i.e. relayphp possible as signatures match
  closely

Performance
-----------
Using the TechEmpower 20 database queries benchmark in 2000
transactions. The new implementation including full BC shows a minimal
overhead, removing the BC it's minimally faster. The new implementation
always uses less memory but more function calls (~20-60 per benchmark
request of ~5000 total).

```
                   trans/sec    memory
1.1                  14.90      ~2.48M
1.1+new-filters+BC   14.84      ~2.45M
1.1+new-filters      15.04      ~2.42M
```

Original Goals
--------------
- PHP7 compatibility
- Plans to flatten hierarchy by removing Object/StaticObject
- Separating into package for easier breakout later

Why a centralized approach?
---------------------------
Previous implementation was already semi-centralized, to make lazy filters work.
This is just the consequent further development of it.

The centralized approach allows further consolidation, there is now just
one place to document and to maintain code at. Duplications of filtering
logic in `Libraries`, `Model`, `Mocker`, `Object`, `StaticObject` have
all been removed.

As closure filters are not stored on an instance anymore certain objects
become easier serializable i.e. `Entity` and `data\Collection`.

Why change signatures?
----------------------
The old filter and chain advancing signatures were more verbose then
needed as they had to be compatible with PHP 5.3.

The new signatures leverage new context binding features (i.e.
availability of `$this`) and drop unneeded parameters. This makes them
more similar to other middleware handlers (like relayphp).

Why is $chain passed as $next?
-------------------------------
This is to hide the implementation details and makes the concept easier
to understand.

Why isn't params part of chain?
-------------------------------
It would need us to expose the full chain object.

Why drop `method()` from chain?
--------------------------------
This would make Chain not context-free and break isolation.

Why not keep the implementation (callback) as part of the chain?
------------------------------------------
It's conceptually wrong to make it part of that.

Open for Discussion
-------------------
The final package name (`aop`) isn't cast in stone, yet. If somebody has a better
name for it, the name can be changed.

Known Limitations
-----------------
Mocker still uses applyFilter on mocked classes, as mocked class names
are hard to guess and to be passed to Filters::apply(CLASS).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,64 @@
   ```
 
 
+- Filters and related classes have been refactored and extracted into the new
+  `lithium\aop` namespace.
+
+  **Shallow hirarchy.** With the new `Filters`, classes that need filter functionality
+  don't need to inherit from `Object`/`StaticObject` anymore and don't need to have
+  special methods or properties defined. This is possible by entirely relying on a central
+  filters manager and using `spl_object_hash()` internally.
+
+  **Simplified filters signature.** By using PHP 5.4's new context binding feature for
+  closures, we were able to simplify the signature of filters - i.e. by dropping the
+  `$self`.
+
+  This - as a sideeffect - reduces the requirement of using `invokeMethod()` to access
+  protected members of the context. `$this` and `static` can be used to access
+  the filtered object. Also makes better stacktraces.
+
+  **Simplified chain advancing.**
+  Instead of advancing the chain via `$chain->next()`, a callable (`$next()`) is used.
+  The advancing function just requires one argument.
+
+  ```php
+  function($self, $params, $chain) { // old
+      $self->invokeMethod('_foo');
+      $params['bar'] = 'baz';
+      return $chain->next($self, $params, $chain);
+  }
+
+  function($params, $next) { // new
+      $this->_foo();
+      $params['bar'] = 'baz';
+      return $next($params);
+  }
+  ```
+
+  **Strict separation of filters and implementation.** The implementation function (the one
+  being filtered) now takes just a single `$params` argument. It doesn't have access to
+  the chain anymore.
+
+  **Everything is lazy.** The new filter manager will now by default apply any filters -
+  for both concrete and static classes - lazily.
+
+  **Better filtering of concrete classes.** It is now possible to apply
+  filters to both all and/or specific instances of a concrete class.
+
+  ```php
+  $foo = new Foo();
+
+  // Executes filter only for `$foo`'s `bar` method.
+  Filters::apply($foo, 'bar', function($params, $next) {});
+
+  // Executes filter for all `Foo`'s `bar` methods.
+  Filters::apply('Foo', 'bar', function($params, $next) {});
+  ```
+
+  **Performance** for filtered methods - especially when there are no filters applied to it -
+  has been improved, so that there is only a minimal penality for making a method
+  filterable.
+
 ### Added
 
 - `Cache` together with the `File` adapter can now be used to store BLOBs. 
@@ -125,16 +183,16 @@
   present in the command class definition. Previously `--no-color` was made available as
   `no-color`. This has been deprecated.
 	
-- The String class has been renamed to `Text` while RNG and hashing functionality 
+- The String class has been renamed to `Text` while RNG and hashing functionality
   have been extracted into `lithium\security\Random` and `lithium\security\Hash`. #1184 (David Persson)
 
-  This is mainly to achieve PHP 7.0 compatibilty as `String` 
-  [will become a reserved name](https://wiki.php.net/rfc/reserve_even_more_types_in_php_7).  
-  
-  Old methods are deprecated but continue to work and redirect to new methods. It wont be 
-  possible to use the old String class with PHP >= 7.0. You must use the new names before 
-  switching to PHP 7.0. 
-  
+  This is mainly to achieve PHP 7.0 compatibilty as `String`
+  [will become a reserved name](https://wiki.php.net/rfc/reserve_even_more_types_in_php_7).
+
+  Old methods are deprecated but continue to work and redirect to new methods. It wont be
+  possible to use the old String class with PHP >= 7.0. You must use the new names before
+  switching to PHP 7.0.
+
   | old | new |
   | --- | --- |
   | `lithium\util\String::hash()` | `lithium\security\Hash::calculate()` |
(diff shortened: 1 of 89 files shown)


jails:
Wow that's an ambitious idea :+1: !
My 2 cents on that:

1. I found that `spl_object_hash()` is a bit slow in practice. So if you are calling a filterable method in a loop (even with no filter defined) you will still need to build the object ID as many times the method is called to check for attached filters. I don't have any precise stats on it and if it could leads to some performance issues. But the workaround I did [here](https://github.com/crysalead/filter) is to use a trait so the look up is done by an extra method call only. It still not perfect because it requires to add `use Filterable;` explicitly and it's also BC break.
2. The fact that's there's not registering/unregistering mechanisms in li3 makes filters difficult to test in practice. For example it's sometime interesting to remove filters for a specific test but being able to restore them on a "tear down". So maybe we can also improve the API on that part.
3. `$params` requires some `compact()`/`extract()` step or requires some additionnal rewriting for making a method filterable. I know it's also BC Break but I would prefer to remove `$params` in flavor of real method parameters. I'll be way easier to make a method filterable. And the fact it won't require any modification of the method content open the door to some Just In Time code patching. So we won't need to explicitly manually refactor a method to be filterable, simply applying a filter to it will make if filterable and the code will be automatically patched on the fly. 

Unfortunately I didn't find any perfect AOP implementation in PHP yet, so I guess we will still need to make some compromises.

davidpersson:
> Wow that's an ambitious idea :+1: !

Thanks, @jails :) 

It's ambitious but necessary. I'm not a friend of monster patches but this was needed in order to extract the implementation. It also shows how interwoven everything is.

> My 2 cents on that:

I've actually looked through your implementation (and others) before this and found many good things, like splitting the class into Filter and Chain, reducing the verbosity of the signatures and more. So thank you for your work there!  

> I found that spl_object_hash() is a bit slow in practice. So if you are calling a filterable method in a loop (even with no filter defined) you will still need to build the object ID as many times the method is called to check for attached filters. I don't have any precise stats on it and if it could leads to some performance issues. 

In my benchmarks the implementation using spl_object_hash doesn't seem particularly slow (see the benchmarks I did above). I think we always need to test the whole implementation. I'll - however - can do some benchmarks and post them here.

Filtering (or any semi-expensive function call) in a heavy loop will always be slow. Our goal should be to identify places where no filtering is needed at all and remove those. This will give us most performance gain.

> But the workaround I did here is to use a trait so the look up is done by an extra method call only. It still not perfect because it requires to add use Filterable; explicitly and it's also BC break.

I wanted a complete isolated and central solution so keeping filters on objects wasn't a possibility. Central because all filters are now lazily applied and lazy because else applying a filter means always loading the target class' file.

> The fact that's there's not registering/unregistering mechanisms in li3 makes filters difficult to test in practice. For example it's sometime interesting to remove filters for a specific test but being able to restore them on a "tear down". So maybe we can also improve the API on that part.

In my opinion filters should be tested in isolation with the class they wrap around mocked. All classes on the other hand should be tested without any filters at all. That's why there is just a "clear" method.

> $params requires some compact()/extract() step or requires some additionnal rewriting for making a method filterable. I know it's also BC Break but I would prefer to remove $params in flavor of real method parameters. I'll be way easier to make a method filterable. 

I also find compact/extract steps not perfect. There was just no way to make this BC compatible. I'd suggest to revisit the topic for the next major version.

> And the fact it won't require any modification of the method content open the door to some Just In Time code patching. So we won't need to explicitly manually refactor a method to be filterable, simply applying a filter to it will make if filterable and the code will be automatically patched on the fly.

JIT code patching is in my opinion too experimental for the core. It also requires more classes to accomplish its task. This is a feature which would perfectly fit into a plugin.

> Unfortunately I didn't find any perfect AOP implementation in PHP yet, so I guess we will still need to make some compromises.

You're right it's never a one fits all solution. But what we have here - a pretty minimalistic approach - is more then good enough.

Once again thanks for your feedback I very much appreciate a lively discussion about this patch.

davidpersson:
Tracked in #1224

davidpersson:
@nateabele It would be great to have this (in the current form or another) in 1.1 to continue our work inside Object and StaticObject. I know that this is a huge patch and no easy to review, but would appreciate if we can move this forward.

nateabele:
Sorry, I don't know how I missed this before. Reviewing now...

shopblocks:
Any luck in getting this merged, it will help in getting the rest of the Object and StaticObject things sorted. Thanks.

nateabele:
BC-compatible and everything. :vader: Impressive. Most impressive.

:+1:

alexbowers:
@nateabele Are we able to merge this soon? I'd like to work on the Object / StaticObject refactoring,but would like the filtering to be taken out first (which this does)

nateabele:
Looks like it needs to be rebased. After that, sure.

davidpersson:
@alexbowers Give me 2 days :)

alexbowers:
@davidpersson Rebase? :D

davidpersson:
@nateabele This is now good to merge, there is just the PHP7 test failing (which we are working on elsewhere and was failing before).

shopblocks:
Should we have the PHP7 as a requirement in travis before we have the tests actually working?

davidpersson:
We'll have them working very soon. So I think this maybe not perfect but OK now.

shopblocks:
Any update on getting these merged @davidpersson

davidpersson:
This is good to merge. Test fail can be ignore see my comment above.

shopblocks:
It looks like a rebase is necessary though since there are conflicts.

davidpersson:
@nateabele Lets merge this. All good tests see above comment.

ericcholis:
Something worth noting for those that are migrating to this new implementation. Since this now uses PHP 5.4's context binding feature for closures, be sure to apply your filter within the proper context if you plan on using `$this` or `static`.

For example, I had a filters.php file in my bootstrap with some common filters. These used $self in various places, which is now deprecated. Also, I had some older filters declared directly in the model, after the class definition. While both of these are convenient for organization, they aren't necessarily the best in practice.

jails:
Imo this issue can be solved by using `Closure::bindTo()` inside `::apply()` like in the following https://github.com/crysalead/filter/blob/bcf97ea0c1f86fc920791fd3b7644e8282e3747b/src/Filters.php#L63-L75

ericcholis:
Excellent @jails, I completely missed that.

PR #1216, opened 2015-10-23
Guarantee strong source in Random::generate(); add sources.

davidpersson:
Add mcrypt_create_iv() as source.
Add PHP7 function random_bytes() source.
Throw exception when no suitable source was found.
Close handle on /dev/urandom.


nateabele:
:+1:

PR #1222, opened 2015-12-13
Add ability to cache BLOBs using file cache adapter

davidpersson:
This introduces a new configuration option to the file cache adapter, which will cause it to return stream handles instead of the value itself. This allows for reading and writing BLOBs. 

The file format remains the same along the file's meta data header.

This feature is introduced to cover the following use cases:

1. You compile a PDF upon user request, an expensive operation is kicked of to compile the PDF. Upon following requests from other users, the PDF should be returned from cache to reduce compile overhead.

2. Your application downloads an instagram feed and stores the data together with images. Each image is downloaded into a temporary file, then a resize process will generate a "version" off it. As the version processing is isolated and knows nothing about the other resize processes the image would need to be re-downloaded for each version.  

I plan to add a BLOB caching section to the caching manual.

### DIFF
--- a/storage/cache/adapter/File.php
+++ b/storage/cache/adapter/File.php
@@ -23,7 +23,7 @@ use lithium\storage\Cache;
  * for single-keys only. Clearing the cache is supported. Real persistence of cached items
  * is provided. Increment/decrement functionality is provided but only in a non-atomic way.
  *
- * This can't handle serialization natively. Scope support is available but not natively.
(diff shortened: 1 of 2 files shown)


nateabele:
Would it be less code if the adapter were a subclass of `File`?

davidpersson:
After a review of the File class I think unfortunately not. There would be too many methods to override as we have our own file format (a small header added to the cache files designating expiry).

nateabele:
Okay, fine with me then. :+1:

PR #1223, opened 2015-12-16
Implement the Filterable Trait

shopblocks:
@davidpersson Can you check this is the sort of thing you are looking for, if so I will continue working on it, otherwise i'll fix this then continue. 

Thanks.

### DIFF
--- a/action/Controller.php
+++ b/action/Controller.php
@@ -8,6 +8,7 @@
 
 namespace lithium\action;
 
+use lithium\core\Filterable;
 use lithium\util\Inflector;
 use lithium\action\DispatchException;
 use lithium\core\Libraries;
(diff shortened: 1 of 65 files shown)


davidpersson:
In principle this would be the way to go. Except filtering support is solved elsewhere (PR). So this particular change might not be a good example. For the time it is OK to name traits the way you did, but we'll need a discussion about the naming (i.e. Filterable vs FilterTrait).

shopblocks:
Okay, so should I continue my work based on that PR, rather than the 1.1 branch?

davidpersson:
If possible basing your work on the 1.1 branch would be preferable. Hopefully we don't get into merge hell when the PR is actually merged. But at this point I don't want to hold you back and can also not guarantee in which form/if the PR will be merged.

davidpersson:
Also lets start one step/change at a time. So I can review your code more easily and help you get it merged.

shopblocks:
> If possible basing your work on the 1.1 branch would be preferable

okay will do.

>  So I can review your code more easily and help you get it merged.

Cool.

I'm still trying to just get the tests themselves to pass locally. 
Is there something that I need to have setup to get the tests to pass locally?

PR #1228, opened 2015-12-16
Plain test output

shopblocks:
Example of usage: 
```bash
bash console/li3 test tests --format=plain |& tee plain.txt
```

Example Output:

before:
```
----
[1;36mTest[0m
----

[0;36mS[0m.......................[0;36mS[0m[0;36mS[0m......[0;36mS[0m[0;36mS[0m[0;36mS[0m.........................
............................................................
..[0;36mS[0m...[0;31mF[0m.....................................................
..[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m[0;36mS[0m.............................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
...............[0;35mE[0m..[0;35mE[0m.........................................
............................................................
............................................................
............................................................
............................................................
................
```

After:

```
----
Test
----

S.......................SS......SSS.........................
............................................................
..S...F.....................................................
..SSSSSSSSSSSSS.............................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
............................................................
...............E..E.........................................
............................................................
............................................................
............................................................
............................................................
................
```


shopblocks:
Also works with verbose mode (where it is more useful)


```
----
Test
----

[     skip] on line  136 in ??::??()
[     pass] on line   38 in lithium\tests\integration\storage\CacheTest::testFileAdapterCacheConfig()
[     pass] on line   42 in lithium\tests\integration\storage\CacheTest::testFileAdapterCacheConfig()
[     pass] on line   57 in lithium\tests\integration\storage\CacheTest::testReadThroughWithStrategies()
[     pass] on line   78 in lithium\tests\integration\storage\CacheTest::testFileAdapterReadThroughWithStrategies()
[     pass] on line   82 in lithium\tests\integration\storage\CacheTest::testFileAdapterReadThroughWithStrategies()
[     pass] on line   97 in lithium\tests\integration\storage\CacheTest::testMultiWriteReadWithStrategies()
[     pass] on line  108 in lithium\tests\integration\storage\CacheTest::testMultiWriteReadWithStrategies()
[     pass] on line  123 in lithium\tests\integration\storage\CacheTest::testMultiWriteReadWithMultipleStrategies()
[     pass] on line  134 in lithium\tests\integration\storage\CacheTest::testMultiWriteReadWithMultipleStrategies()
[     pass] on line  150 in lithium\tests\integration\storage\CacheTest::testFileAdapterWrite()
[     pass] on line  155 in lithium\tests\integration\storage\CacheTest::testFileAdapterWrite()
[     pass] on line  158 in lithium\tests\integration\storage\CacheTest::testFileAdapterWrite()
[     pass] on line  159 in lithium\tests\integration\storage\CacheTest::testFileAdapterWrite()
[     pass] on line  177 in lithium\tests\integration\storage\CacheTest::testFileAdapterWithStrategies()
[     pass] on line  183 in lithium\tests\integration\storage\CacheTest::testFileAdapterWithStrategies()
[     pass] on line  186 in lithium\tests\integration\storage\CacheTest::testFileAdapterWithStrategies()
[     pass] on line  189 in lithium\tests\integration\storage\CacheTest::testFileAdapterWithStrategies()
[     pass] on line  190 in lithium\tests\integration\storage\CacheTest::testFileAdapterWithStrategies()
[     pass] on line  208 in lithium\tests\integration\storage\CacheTest::testFileAdapterMultipleStrategies()
[     pass] on line  214 in lithium\tests\integration\storage\CacheTest::testFileAdapterMultipleStrategies()
[     pass] on line  217 in lithium\tests\integration\storage\CacheTest::testFileAdapterMultipleStrategies()
[     pass] on line  220 in lithium\tests\integration\storage\CacheTest::testFileAdapterMultipleStrategies()
[     pass] on line  221 in lithium\tests\integration\storage\CacheTest::testFileAdapterMultipleStrategies()
[     skip] on line  136 in ??::??()
[     skip] on line  136 in ??::??()
[     pass] on line   91 in lithium\tests\integration\storage\SessionTest::testMemoryReadWriteDelete()
[     pass] on line   92 in lithium\tests\integration\storage\SessionTest::testMemoryReadWriteDelete()
[     pass] on line   93 in lithium\tests\integration\storage\SessionTest::testMemoryReadWriteDelete()
[     pass] on line   94 in lithium\tests\integration\storage\SessionTest::testMemoryReadWriteDelete()
[     pass] on line   95 in lithium\tests\integration\storage\SessionTest::testMemoryReadWriteDelete()
[     pass] on line   96 in lithium\tests\integration\storage\SessionTest::testMemoryReadWriteDelete()
[     skip] on line  136 in ??::??()
[     skip] on line  136 in ??::??()
[     skip] on line  136 in ??::??()
[     pass] on line   49 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testEnabled()
[     pass] on line   60 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testSimpleWrite()
[     pass] on line   61 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testSimpleWrite()
[     pass] on line   64 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testSimpleWrite()
[     pass] on line   74 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testSimpleWrite()
[     pass] on line   78 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testSimpleWrite()
[     pass] on line   81 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testSimpleWrite()
[     pass] on line   92 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteExpiryDefault()
[     pass] on line   96 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteExpiryDefault()
[     pass] on line   99 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteExpiryDefault()
[     pass] on line  109 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteNoExpiry()
[     pass] on line  112 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteNoExpiry()
[     pass] on line  120 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteNoExpiry()
[     pass] on line  123 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteNoExpiry()
[     pass] on line  131 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteNoExpiry()
[     pass] on line  134 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteNoExpiry()
[     pass] on line  145 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteExpiryExpires()
[     pass] on line  156 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteExpiryExpires()
[     pass] on line  165 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteExpiryTtl()
[     pass] on line  176 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteExpiryTtl()
[     pass] on line  189 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteMulti()
[     pass] on line  193 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteMulti()
[     pass] on line  197 in lithium\tests\integration\storage\cache\adapter\MemcacheTest::testWriteMulti()
```

shopblocks:
Issue stemming from developing this: https://github.com/UnionOfRAD/lithium/issues/1229

davidpersson:
For my taste this is too much code duplication from the already present text mode. I'd rather suggest to detect piping using `posix_isatty` automatically (if module is present) or disabling ansi escaped code manually. This could become a global console feature.

shopblocks:
I think i prefer the --no-color option. I'll look into putting that in then.

nateabele:
Agreed with @davidpersson.

shopblocks:
Since this should be a user specified thing, I think that we should have the flag. I'll work on that later.

davidpersson:
No objections. Keep the changes as minimal as possible.

shopblocks:
Will do.

shopblocks:
@davidpersson @nateabele Thoughts?

shopblocks:
I have noticed some areas that aren't covered. I'll expand on it.

console/Command.php
```diff
@@ -57,6 +57,13 @@ class Command extends \lithium\core\Object {
 	public $plain = false;
 
 	/**
+	 * Shows text output with color or without
+	 *
+	 * @var boolean
+	 */
+	public $no_color = false;
```
davidpersson: Should be camelCased.

console/command/Test.php
```diff
@@ -59,6 +59,13 @@ class Test extends \lithium\console\Command {
 	public $verbose = false;
 
 	/**
+	 * Shows text output with color or without
+	 *
+	 * @var boolean
+	 */
+	public $no_color = false;
```
davidpersson: Isn't this inherited from Command already?

shopblocks: Yes it is. Fixed.

console/Command.php
shopblocks: Fixed.

shopblocks:
Changed the way that this all works, so that Response->styles() instead returns no ansi codes, so that it covers all situations, rather than the first attempt which was on a per-section basis

console/Command.php
```diff
@@ -109,17 +116,23 @@ protected function _init() {
...
 
 		foreach ($params as $key => $param) {
+			if($key == 'no-color') {
+				$key = 'noColor';
+			}
+
 			$this->{$key} = $param;
 		}
+
+		$this->response = $this->_config['response'];
```
shopblocks: This has been reordered, so that all request sections are dealt with before the response, this is because the response depends on the noColor section of the request.

The alternative is to process noColor first, and then process the rest afterwards.

This made more sense, and since none of the rest of the response depended on the request, it seems fine to do this way.

console/command/Test.php
shopblocks: Reverted this, because it is necessary for the 'help' command. It doesn't seem to take into account the inherited properties.

I am now following the setup for the rest of the properties (plain is defined in both Command and Test)

shopblocks:
When you are happy with the merge let me know, and i'll squash it all down to 1 commit.

console/Response.php
```diff
@@ -126,6 +136,29 @@ public function styles($styles = array()) {
...
+				'command' => "",
+				'error'   => "",
+				'success' => "",
+				'bold'    => "",
+			);
+
+			$styles = false;
+		}
+
 		if ($styles === false) {
```
davidpersson: As far as I can see, you could remove the block above by changing this line to: 
`if ($styles === false || $this->noColor)` 

as a bonus round line 167-169 could also be removed by changing it to:

`if ($styles === false || $this->noColor || strtoupper(substr(PHP_OS, 0, 3)) === 'WIN')`

console/command/Test.php
```diff
@@ -67,6 +67,13 @@ class Test extends \lithium\console\Command {
 	public $plain = false;
 
 	/**
+	 * Shows text output with color or without
+	 *
+	 * @var boolean
+	 */
+	public $noColor = false;
```
davidpersson: This should be inherited?

console/command/Test.php
```diff
@@ -90,7 +97,8 @@ protected function _init() {
 					$command->header('Test');
 					$command->out(null, 1);
 				}
-				$colorize = function($result) {
+
+				$colorize = function ($result) {
```
davidpersson: Mistakenly adding whitespace here.

console/Command.php
```diff
@@ -109,17 +116,23 @@ protected function _init() {
...
-		$this->response = $this->_config['response'];
 
-		if (!is_object($this->response)) {
-			$this->response = $this->_instance('response', $this->response);
-		}
 		$default = array('command' => null, 'action' => null, 'args' => null);
 		$params = array_diff_key((array) $this->request->params, $default);
 
 		foreach ($params as $key => $param) {
+			if($key == 'no-color') {
```
davidpersson: Isn't no-color mapped automatically to noColor? We might to change the console routing if slugged multi word args are not possible.

console/command/Test.php
davidpersson: Okay, strange why this is.

console/command/Test.php
shopblocks: I mentioned that I reverted this, because the `bash console/li3 help test` command does not take into account inherited properties. This is done also with the `plain` flag which already existed.

console/command/Test.php
shopblocks: Fixed.

console/Command.php
shopblocks: image

image


Multi words are not automatically mapped to noColor at the moment no.

console/Response.php
shopblocks: Fixed.

console/command/Test.php
davidpersson: Yes, sorry for this. I saw your other comment a bit too late.

shopblocks: No problem. Its strange, and perhaps should be looked into. but for now, duplicating a property is not an issue I don't think.

console/Command.php
shopblocks: @davidpersson I have updated the router to work as expected.

shopblocks:
This is failing on a test currently for GNU style long options, because the routing has changed to be from 'foo-bar' to 'fooBar'

Changing the test to match 'fooBar' passes, but this is now a backwards incompatible change. 

What do you think is a good idea here?

1) I revert that code, and file a new PR and issue on this topic

2) Keep the code as is, and fix the test. List it as a backwards incompatible change

3) Set both options, both 'foo-bar' and 'fooBar'

shopblocks:
I have decided to revert the router changes. 

I will submit them in a different PR

console/Command.php
davidpersson: Seems good to me.

console/Response.php
```diff
@@ -126,14 +136,12 @@ public function styles($styles = array()) {
...
+
+		if ($styles === false || $this->noColor || strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
 			return array_combine(array_keys($defaults), array_pad(array(), count($defaults), null));
 		}
 		$styles += $defaults;
 
-		if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
-			return $this->styles(false);
-		}
 		return $styles;
```
davidpersson: Save more lines by making the last one `return $styles + $defaults`

alexbowers: Done

console/Command.php
```diff
@@ -57,6 +57,13 @@ class Command extends \lithium\core\Object {
 	public $plain = false;
 
 	/**
+	 * Shows text output with color or without
```
davidpersson: Might be reworded as `Disables color output. Useful when piping command output into other commands.`. Also the text needs to be the same as in `Help::$noColor`

shopblocks: Done.

shopblocks:
@davidpersson I think this is ready for merging

davidpersson:
I'm working on one thing: Inheritance of command properties. That's missing and that's why `$noColor` needs to be redefined in i.e. `Test` but it would also need to be redefined in all other commands.

shopblocks:
Okay, cool.

davidpersson:
Along the way I'm discovering more problems. We might need to take a different route with this. Will report later.

shopblocks:
Is the issue with plain test output, or with the way no-color was implemented?

shopblocks:
Or with commands in general?

shopblocks:
http://php.net/manual/en/reflectionclass.getdefaultproperties.php I think this may be useful.

davidpersson:
So...

inherited command line options (i.e. `--silent`) have always been available to all commands, they were just not shown by the help command. This has been fixed in e10b977539ad64c94fd7691105504d6b77ecca56. 

Along the way I've discovered we already had an option to disable color output (`--plain`). Through the removal of the redundant plain property in the test command this becomes more obvious.
https://github.com/UnionOfRAD/lithium/commit/e10b977539ad64c94fd7691105504d6b77ecca56#diff-acd2b576d08bdb2acab1cb5e3cc8958aL62

I've also took the chance - as you already did some refactoring work - to improve the colorization control and move it into the response object. bd45e73b7c6295609ad405ed18af0eb71715a200 I've credited you in the commit message.

tl;dr; Use `li3 test --plain tests | tee ...` to get non colored output.

shopblocks:
Its probably worth noting this strips out decoration too, such as headers.

shopblocks:
@davidpersson Major issue with this. The decoration is actually required.

It does not give the overview at the end, even with verbose mode on.

Previous examples of the code I had did do this.

davidpersson:
Okay than lets not repurpose the `--plain` flag in the test command, but introduce a new one i.e. `--only-results`?

shopblocks:
For backwards compatibility, having --plain only do the dots etc (colourless now), makes sense so that nobody has unexpected changes.

Instead of --only-results, which would mean that --plain is deviated from what it used to do, we should instead now respect --verbose to give output, and perhaps have another flag also that shows results

That to me, should be --show-results

This would then allow someone to have --plain --show-results to have it all dots etc with results at the end, or --plain --verbose to get the verbose with results at the end.

I'm not sure whether having verbose do results is the best thing, perhaps if we have to add --show-results then we have verbose do what it currently does in terms of just test names etc, and then --show-results is needed to get results as well.

davidpersson:
Some clarification: There would be an option that would just show the dots (or with verbose enabled, just the assertions). I think it's okay to make this non BC. It's a minor change in a part of the non-critical developer tooling options.

shopblocks:
Can you explain what the default behaviour would be then and the new flags

davidpersson:
The default behavior without any options, stays as is.
```
# Show only the "dots" no header/footer
test --just-assertions 

# Shows only the assertions but verbose and no header/footer. 
test --just-assertions --verbose 

# Show everything but no colorization
test --plain
```

shopblocks:
So default behaviour with --plain and test --just-assertions are going to be the same then?

davidpersson:
--plain switches colorization on/off, --just-assertions switches decoration on/off. Both can be combined. The original behavior of --plain changes and will not include --just-assertions.

shopblocks:
Okay, are you going to be working in this?

davidpersson:
Yes :)

shopblocks:
Okay cool.
 I'll delete my branch again then.

Do you have the chance to look into the PHP 7 String merge i just made, want your input on that before I progress further

davidpersson:
Will have a look at it -- thanks!

davidpersson:
Handled in 342f64300d8c0a7596bb63d529f4327ba798c304

shopblocks:
Can we get `--just-assertions` to output a newline at the end, so that it doesn't make the output look weird?

davidpersson:
Looks good to me. What do you mean exactly? 
bildschirmfoto 2015-12-17 um 16 16 36

shopblocks:
image

shopblocks:
It does the same on just `--just-assertions` as well.

shopblocks:
Does the same in iTerm2 and Terminal on Mac OS X. So it isn't an environment thing as far as i'm aware.

Issue #1231, opened 2015-12-16
Routing for console does not convert dashes to camelcase


shopblocks:
When passing in a flag for --no-color, it does not convert to noColor.

There is a test that checks and verifies that 'foo-bar' stays as 'foo-bar', however this means that to meet current conventions for variable names (camelcased), individual checks have to be made for flags.

Example can be seen [here](https://github.com/shopblocks/lithium/blob/feature/plain-test-output/console/Command.php#L124-L126).

Changing the current behaviour will break backwards compatibility, so I reverted a change I made to #1228 pull request.

https://github.com/shopblocks/lithium/blob/feature/plain-test-output/tests/cases/console/RouterTest.php#L88 is the test that will be broken if this change is made.

PR #1232, opened 2015-12-16
CamelCase multi word command args

shopblocks:
Fixes https://github.com/UnionOfRAD/lithium/issues/1231


console/Router.php
```diff
@@ -34,11 +36,13 @@ public static function parse($request = null) {
 			while ($args) {
 				$arg = array_shift($args);
 				if (preg_match('/^-(?P<key>[a-zA-Z0-9])$/i', $arg, $match)) {
-					$params[$match['key']] = true;
+					$key = Inflector::camelize(Inflector::classify($match['key']), false);
```
davidpersson: I think we can drop ::classify() here?

console/Router.php
```diff
@@ -34,11 +36,13 @@ public static function parse($request = null) {
...
+					$params[$key] = true;
 					continue;
 				}
 				if (preg_match('/^--(?P<key>[a-z0-9-]+)(?:=(?P<val>.+))?$/i', $arg, $match)) {
-					$params[$match['key']] = !isset($match['val']) ? true : $match['val'];
+					$key = Inflector::camelize(Inflector::classify($match['key']), false);
```
davidpersson: ... and here

davidpersson:
This can be made BC, when we keep the uninflected AND the original key.

console/Router.php
shopblocks: You are correct.

shopblocks:
> This can be made BC

I'll make the changes and see what you think

shopblocks:
@davidpersson Should the test be duplicated also to test for both occurrences?

nateabele:
Yeah, the test should assert that both keys are present, so add a line instead of replacing the existing one. Thanks.

alexbowers:
Done.

console/Router.php
```diff
@@ -53,4 +59,4 @@ public static function parse($request = null) {
 	}
 }
 
-?>
```
davidpersson: There should be no newline at the end of the file.

console/Router.php
```diff
@@ -8,6 +8,8 @@
 
 namespace lithium\console;
 
+use lithium\util\Inflector;
+
 /**
  * The `Router` class uses an instance of `lithium\console\Request`, which represents an incoming
  * command-line invocation, to parse the correct command, and sub-command(s) and parameters, which
```
davidpersson: It would be great if it could be mentioned, that "foo-bar" in results is deprecated in favor of "fooBar". This will later help us remove deprecated features.

console/Router.php
shopblocks: This was the online github editor that did that.

I'll fix locally. Thanks.

console/Router.php
shopblocks: Mentioned in the comments?

console/Router.php
```diff
@@ -34,10 +36,14 @@ public static function parse($request = null) {
 			while ($args) {
 				$arg = array_shift($args);
 				if (preg_match('/^-(?P<key>[a-zA-Z0-9])$/i', $arg, $match)) {
+					$key = Inflector::camelize($match['key'], false);
+					$params[$key] = true;
 					$params[$match['key']] = true;
```
davidpersson: If possible make this one line (only if the same ca be done for line 46&47 too without exceeding line length limit -- for symmetry). 
```$params[$key] = $params[$match['key'] = true```

File: console/Router.php (hunk 4 above)
Comment (davidpersson) [inline console/Router.php]: Mentioned in the docblock for `parse()`.

Comment (shopblocks) [inline console/Router.php]:
```php
/**
	 * Parse incoming request from console. Short and long (GNU-style) options
	 * in the form of `-f`, `--foo`, `--foo-bar` and `--foo=bar` are parsed.
	 * XF68-style long options (i.e. `-foo`) are not supported but support
	 * can be added by extending this class.
	 *
	 * If passing through `--foo-bar` this previously (pre 1.1) resulted in 
	 * the option `foo-bar` being available. This has been deprecated in favor
	 * of `fooBar` being created
	 *
	 * @param \lithium\console\Request $request
	 * @return array $params
	 */
```

Happy with that as the new comment?

console/Router.php
shopblocks: What is the line limit?

console/Router.php
davidpersson: Add a dot at the end of the sentence and all is perfect :)

console/Router.php
davidpersson: Good question. From LSR-0:

> Lines must not exceed a limit of 100 characters. The soft limit is 80 characters. Lines should not have any trailing whitespace characters.

console/Router.php
shopblocks: Done

console/Router.php
shopblocks: This limit is including indentation?

shopblocks: It comes to 91 with indentation according to an online counter. Pushed changes.

davidpersson: Yes, including indentation. (I will add this clarification to the spec.)

shopblocks: I also counted a tab as 1, should it be 4?

davidpersson:
Looks nice. When you're fine I can merge this.

shopblocks:
I'm happy with this as is.

shopblocks:
Are you happy with 8 commits, or would you prefer me to flatten it into 1?

davidpersson:
Thanks, I'll squash them myself.

shopblocks:
Okay. great. In that case, im happy with this as is.

shopblocks:
With this being merged in, should I make the change to the https://github.com/UnionOfRAD/lithium/pull/1228 to no longer have the special check in place for 'no-color' https://github.com/shopblocks/lithium/blob/feature/plain-test-output/console/Command.php#L124-L126

davidpersson:
> With this being merged in, should I make the change to #1228 to no longer have the special check in place for 'no-color'

Yes :) and thanks for the patch!

shopblocks:
Is it safe for me to delete the branch locally now?

davidpersson:
Yes, go ahead.

PR #1233, opened 2015-12-17
Throw exception on null template

shopblocks:


shopblocks:
Im not sure whether there should be a test for this.

As a quick outline for what it does.

If you create a new format for tests, and then run the tests, when it gets to the 'results' section, it throws a warning saying you cannot include an  empty file (the template is null, so it tries `include null`)

This simply helps the developer find an issue, rather than trying to track it down like I did yesterday.

davidpersson:
This is OK. Thank you for working on the patch.

(I'll fix the one minor coding standard when I'll merge this.)

shopblocks:
Whats the code standard violation?

test/Report.php
```diff
@@ -229,6 +230,11 @@ public function render($template, $data = array()) {
 		$template = Libraries::locate('test.templates', $template, array(
 			'filter' => false, 'type' => 'file', 'suffix' => ".{$config['format']}.php"
 		));
+
+		if($template === null) {
```
davidpersson: Between the `if` and the `(` one whitespace is needed to be in line with the CS.

shopblocks: Ah okay. I can fix that if you like.

Also, do you want the commits to be flattened or are you happy with the several commits being merged?

davidpersson:
Merged in c87c87d3ea3f4616d19be17080137fdab6e2b998. Thanks for the patch.

Issue #1224, opened 2015-12-16
PHP 7 Compatibility


davidpersson:
This is a meta bug, tracking and organizing the effort in making 1.1 PHP 7 compatible.

- [x] no PHP internal deprecation warnings
- [x] fix string class loading
- [x] Fix segfault in tests #1236
- [x] Fix #1240 
- [x] all tests green

shopblocks:
While running the tests locally, I have an issue which says: 

```
Failed assertion assertEqual.
 File    : /var/www/vhosts/lithium/tests/cases/net/http/MediaTest.php
 Class   : lithium\tests\cases\net\http\MediaTest
 Method  : testMediaTypes()
 Line    : 39
 ________
trace: [0]
expected: array (
  0 => 'html',
  1 => 'htm',
  2 => 'form',
  3 => 'json',
  4 => 'rss',
  5 => 'atom',
  6 => 'css',
  7 => 'js',
  8 => 'text',
  9 => 'txt',
  10 => 'xml',
)
result: array (
  0 => 'view-integration-test',
  1 => 'html',
  2 => 'htm',
  3 => 'form',
  4 => 'json',
  5 => 'rss',
  6 => 'atom',
  7 => 'css',
  8 => 'js',
  9 => 'text',
  10 => 'txt',
  11 => 'xml',
)
```

Do you have any idea why this would be failing? 

Is there a specific setup I need to test this properly.

I currently just have the repository cloned and run `bash console/li3 test tests`

davidpersson:
This seems to be another issue unrelated to compatibility. Your best bet is to open a new ticket for this.

shopblocks:
issue opened https://github.com/UnionOfRAD/lithium/issues/1225

shopblocks:
I have found the cause of the issue for the `PHP Fatal error:  Cannot use 'String' as class name as it is reserved in /var/www/vhosts/lithium/util/String.php on line 27` issue when running tests.

It is caused by the AffectedTest.php test running,

it runs `Affected::apply()`, which then runs `self::_affected($test->subject())`

This then runs a `Libraries::find(true, compact('exclude') + array('recursive' => true));`

```php
Array
(
    [0] => lithium\util\Collection
    [1] => lithium\util\Inflector
    [2] => lithium\util\Set
    [3] => lithium\util\String
    [4] => lithium\util\Text
    [5] => lithium\util\Validator
    [6] => lithium\util\collection\Filters
[...]

PR #1234, opened 2015-12-17
Dont load the String class if you are in PHP 7+

shopblocks:


shopblocks:
Locally, I get a segmentation fault, being caused by the `testDeprecatedInit()` method in `LibrariesTest.php`

It is calling `MockInitMethod::li3()`.

If i place a `trigger_error('Deprecated', E_USER_DEPRECATED);` in there, it passes fine, without any segmentation faults.

Not sure what the cause of all of this is, but it perhaps needs looking into a bit deeper. May be an issue with the PPA im using for PHP 7, but its the best one available for Ubuntu 14.04 64-bit.

shopblocks:
@davidpersson Are you able to explain what this test should be doing?

```php
public function testDeprectatedInit() {
		$this->assertException("/Deprecated/", function() {
			MockInitMethod::li3();
		});
	}
```

It is causing a segmentation fault in my version, but i can't see what would be wrong. Thanks.

nateabele:
This seems like a really hacky approach. Where is `String` getting loaded from.

shopblocks:
@nateabele Its being loaded because it calls a:

```php
$classes = Libraries::find(true, compact('exclude') + array('recursive' => true));
```

in Affected.php:84.

shopblocks:
I agree to it being hacky, but it is all I could get to work, either that or delete the file outright.

shopblocks:
@nateabele You can see some more details in this thread: https://github.com/UnionOfRAD/lithium/issues/1224

nateabele:
Why not have `Affected` filter out deprecated classes by default?

shopblocks:
This would still cause issues anywhere else that does Libraries::find(true).

This way, it will never cause errors because it will never be loaded.

alexbowers:
Alternative implementation @nateabele https://github.com/UnionOfRAD/lithium/pull/1238

PR #1238, opened 2015-12-17
Fix PHP 7 Fatal Error when including String.php

alexbowers:


davidpersson:
Can't we put the check into the String.php file itself? Or does the problem occur when the file is parsed? In that case another option might be to use the new catchable engine exceptions (https://wiki.php.net/rfc/engine_exceptions_for_php7).

```
if (PHP >= 7) {
    return;
}
```

alexbowers:
I honestly didnt even think of that, i'll try it this evening when I get the chance, unless somebody else wants to give it a go.

alexbowers:
@davidpersson Just gave it  a quick go actually, it doesn't work.

image

alexbowers:
As for the catchable interface, since 1.1 isn't PHP 7 exclusive (as far as i'm aware?), should we be using this feature? Would it not cause parse errors in 5.6 etc.

jails:
This should work imo:
```php
class StringDeprecated {
    ...
}
class_alias('StringDeprecated', 'String');
```

alexbowers:
@jails Nope. 
image

jails:
@alexbowers sorry I forgot the `if (PHP_VERSION_ID < 70000)` part.

alexbowers:
Ah okay, i see what you mean then

alexbowers:
Yep, that seems good.

alexbowers:
Changes pushed

util/String.php
```diff
@@ -19,12 +19,16 @@
 $message .= "class with PHP >=7.0.";
 trigger_error($message, E_USER_DEPRECATED);
 
+if(PHP_VERSION_ID < 70000) {
```
davidpersson: There is just one space missing here.

alexbowers: Fixed

alexbowers:
I assume you dont want all of the comments regarding String::* to be replaced in that class?

davidpersson:
Good catch. But I'd say no we should not do this. Really "Text" should be used and the documentation there is important.

alexbowers:
okay. there is also an unused `use lithium\util\Text` at the top. Do we want this removed or does it not matter?

davidpersson:
It is used actually as far as I can see.

alexbowers:
Weird. PHP storm being weird then.

alexbowers:
Ah, the reason is because they're in the same namespace, no need to 'use' it.

davidpersson:
Yeah, but it's good to be explicit.

davidpersson:
If we're good here, please squash the commits and give it a nice commit message. http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

nateabele:
:+1:

alexbowers:
@davidpersson @nateabele Done.

PR #1253, opened 2016-01-03
2016

Filisko:


alexbowers:
These should probably be in 1 PR?

#1252

alexbowers:
Also, many occurrences are missed. Such as every single file.

davidpersson:
Hi @Filisko, please update this PR with a more complete patchset, feel free to reopen then.

alexbowers:
@davidpersson It also brings  the question of should this be an update?

To me, a license and copyright is from the date it was started, not the current date.

These files were all copyright (probably 2009? but has been updated to 2015), but is that correct?

Should they not be the date when it was copywritten (ie, when it was written).

davidpersson:
Yeah I was thinking about changing our policy here too. But we would need to discuss first. @Filisko until we reach consensus on this topic we need to put this onhold.

alexbowers:
@davidpersson Where can I reach you to discuss possible features for Lithium btw? Is there a slack channel or anything that these discussions can be made on.

alexbowers:
RFC Opened for policy change: #1254

davidpersson:
@alexbowers You can mail me (nperson@gmx.de), than we can chat on IRC (there's a li3-core channel on freenode). Looking forward.

Issue #1178, opened 2015-03-25
Model::save() should only save updated fields


davidpersson:
When saving an entity, just the updated fields should get saved (incremental update).

PR #1121 implemented this feature but it needed to be reverted in 100cd404585fa5660707a92d43edee87d633f403 as a bug in the whitelisting feature couldn't be resolved at that time (#1163).

Prior work for this feature can be found in:
https://github.com/UnionOfRAD/lithium/tree/feature/incremental-db-update

It contains the original patch and test cases as well as some tests.

It would be great to pick up the work and bring this important feature live.

davidpersson:
Implemented in 8b9a42701de3cf407e623a1f827510d11db97c6e.

PR #1262, opened 2016-01-24
Saving data when they are changed

koushki:
Hey guys, I'm coming back :)

As mentioned in #1178 issue

### DIFF
--- a/tests/integration/data/DatabaseTest.php
+++ b/tests/integration/data/DatabaseTest.php
@@ -483,6 +483,77 @@ class DatabaseTest extends \lithium\tests\integration\data\Base {
 		));
 		$this->assertEqual(2, $results->count());
 	}
+
+	public function testUpdateWithMultiThread() {
+		$thread1 = Images::first(1);
+		$thread2 = Images::first(1);
+		
+		$thread1->image = 'tmp';
+		$thread2->title = 'tmp';
+
+		$thread2->save();
+		$thread1->save();
+
+		$this->assertEqual('tmp', Images::first(1)->title);
+	}
+
+	public function testUpdateWithoutFieldsChanged() {
+		$image = Images::first(1);
+		$title = $image->title;
+		$image->title = $title;
+		$this->assertTrue($image->save());
+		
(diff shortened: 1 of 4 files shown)


davidpersson:
Thanks for looking into the issue again. This looks good, especially as not much code is added. Before we can merge, we'll need more tests to make sure this doesn't break something. 

We will need tests for more edge cases.

koushki:
Ok , would you like to me for add more testcases ?

davidpersson:
@koushki Yeah, this would be great.

koushki:
@davidpersson I added more integration test case but i'm not sure which kind of test case were you mean. 
If you think we can add more test case please give me more idea for write them .

davidpersson:
Thanks for adding more cases. I was thinking about edge cases in particular. Off the top of my head:

* Saving all fields updated
* Saving no fields updated
* Saving just a few updated
* Saving non-existing field updated
* Saving using whitelist (all fields updated, none updated, some missing)

tests/integration/data/DatabaseTest.php
```diff
@@ -483,6 +483,33 @@ public function testSubqueryWithMultipleRecords() {
 		));
 		$this->assertEqual(2, $results->count());
 	}
+
+	public function testUpdateWithMultiThread() {
```
davidpersson: We cannot test concurrency reliably/simply, so this test can be removed.

koushki: no it isn't parallelism, just the name is multiThread .

in this test case we have two different object that affect on each other.

davidpersson: Aaah, alrighty. Missed that.

koushki:
@davidpersson Please check again , I added more test cases as you mentioned

davidpersson:
Merged in 8b9a42701de3cf407e623a1f827510d11db97c6e. Sorry that the meta commit info got lost in that merge. Nonetheless thanks for the good work! Let's keep an eye on this.

Issue #1254, opened 2016-01-03
[RFC] Policy change on Copyright Notices


alexbowers:
Currently, the copyright on all files is updated each year, so that we would have it set to the current year.

This seems strange, since licenses and copyright is from when the work was first produced and copy-written .

Since this is a change in policy an RFC is opened.

Filisko:
Well... I don't know too much about this, but have a look to GitHub's, why do you think that they update it?

I think that we should always update it, just in case.

davidpersson:
@Filisko A website's footer (like GitHub's) is something different. Usually companies like to indicate that the website is current.

Go and Apache Mesos don't update the Copyright year. The copyright year is determined by when the file was created. The reason behind this is, that given a project isn't updated for over a year, a year update commit must still occur. As each commit would require a (patch) release this does not make much sense.

https://github.com/golang/go/blob/master/src/os/dir_unix.go

davidpersson:
At this point we can't go back to my suggestion in the last comment. Let's keep it as is.

PR #1263, opened 2016-01-25
Happy 2016

mohnjatthews:
Simply changing the copyright year from 2015 to 2016.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 #
 # Lithium: the most rad php framework
 #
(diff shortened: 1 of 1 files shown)


shopblocks:
https://github.com/UnionOfRAD/lithium/issues/1254

davidpersson:
@mohnjatthews There are many more occurrences. We are also in the process of determining if we want to update the year at all, each time a new one arrives.

mohnjatthews:
Oops, my bad. Didn't see the previous pull request. Won't happen again.

davidpersson:
nvm

Issue #1210, opened 2015-09-04
Null resource when trying to perform raw query


alexbowers-tecmark:
I am trying to perform a raw query to set a session variable in mysql; using the following code:

```php
 Products::connection()->read('SET SESSION group_concat_max_len = 1000000;');
```

This previously worked fine, however on the new 1.0.0-RC1 branch, this is now throwing the following error: 

```
Fatal error: Call to a member function columnCount() on null in /var/www/vhosts/project/libraries/lithium/data/source/Database.php on line 834
```

The object being performed on is: 
```
lithium\data\source\database\adapter\pdo\Result Object
(
    [_iterator:protected] => 0
    [_current:protected] => 
    [_valid:protected] => 
    [_key:protected] => 
    [_resource:protected] => 
    [_autoConfig:protected] => Array
        (
            [0] => resource
        )

    [_buffer:protected] => Array
        (
        )

    [_config:protected] => Array
        (
            [resource] => PDOStatement Object
                (
                    [queryString] => SET SESSION group_concat_max_len = 1000000;
                )

            [init] => 1
        )

    [_methodFilters:protected] => Array
        (
        )

)
```

davidpersson:
From which commit or version have you been updating to 1.0.0-rc1?

davidpersson:
Thank you for your bug report. Should be fixed by 3fa37eb, please test and verify.

alexbowers-tecmark:
Seems fine.

PR #1265, opened 2016-02-03
columnCount() of null fixed

shopblocks:
This is an extension of a previous bug https://github.com/UnionOfRAD/lithium/issues/1210 which didn't completely fix the issue. 

To test this locally, we had in a controller method against mysql ~5.6

```
mysql  Ver 14.14 Distrib 5.6.27, for debian-linux-gnu (x86_64) using  EditLine wrapper
```

Running this code: 

```php
Model::connection()->read('SET SESSION group_concat_max_len = 1000000;');
```

throws a fatal error

```
Fatal error: Call to a member function columnCount() on null in libraries/lithium/data/source/Database.php on line 835
```

We were unable to create a regression test to prove this, and it appears to be a false-positive pass for the previous regression test added https://github.com/UnionOfRAD/lithium/commit/3fa37ebbe5a5f673c39143e3f3a5a0870771f191 due to it being a mocked connection. 

Running it against the actual model which hits mysql throws this fatal error.

We have not tested against SQLite3 or any other storage type.

@davidpersson

### DIFF
--- a/data/source/Database.php
+++ b/data/source/Database.php
@@ -828,7 +828,7 @@ abstract class Database extends \lithium\data\Source {
 		}
 		$result = array();
 
-		if (!$resource) {
+		if (!$resource || !$resource->resource()) {
 			return $result;
 		}
 		$count = $resource->resource()->columnCount();


nateabele:
Seems fine, but it looks like your editor added a trailing newline.

@davidpersson Do you think we need a test for this?

shopblocks:
Thats github editor that did that. I tried to fix it within githubs editor, and again that didnt work.

I'll clone it locally to fix and push the fix.

nateabele:
Thanks.

shopblocks:
Done. Still can't find a regression test to prove this works though.

shopblocks:
Is it perhaps the mock connection that is not mocking it accurately?

shopblocks:
I'll also squash my 3 commits down to 1 commit

davidpersson:
Please also rebase onto 1.0 branch and merge there.

davidpersson:
Can you add a test for this, too?

alexbowers:
Which branches ? 

Just master and 1.0.

Or do i need to do 1.1 and others too?

As mentioned at the top, I couldn't figure out what regression test could cover this, since it (seems) to require a valid mysql connection.

alexbowers:
The old regression test shows the issue if it was using Model::connections()->read('...') instead of $this->_db->read('...')

davidpersson:
Just 1.0. We'll merge "up" into i.e. 1.1 later.

alexbowers:
Okay. I'll do that tomorrow when I'm back in work

davidpersson:
Then I'd say we should add the variation to the test case.

alexbowers:
I had a problem with using Model::connection() though in that there is no 'default' connection. and I didn't have time to investigate how to get around that. Any solution you know of?

davidpersson:
The integration test case (tests/integration/data/Database.php) has an active connection try to use that.

shopblocks:
Sent. https://github.com/UnionOfRAD/lithium/pull/1266

davidpersson:
Closing in favor of #1266.

davidpersson:
Fixed in 191d29589911fba1517e163eb72bacdf2f948e5c

alexbowers:
Thanks :)

davidpersson:
Thank you :)

PR #1266, opened 2016-02-04
columnCount() of null fixed

shopblocks:


shopblocks:
@davidpersson We can't get the integration tests to run for mysql locally it skips them because of a missing 'test' connection.

shopblocks:
Managed to figure it out. It had to be part of an actual application, and couldn't be ran independently. If you remove our fix, you receive the fatal error. With the fix, it passes.

tests/integration/data/source/database/adapter/MySqlTest.php
```diff
@@ -353,6 +353,18 @@ public function testDefaultValues() {
...
+
+	/**
+	 * Verifies that setting options using a raw SQL string works, when
+	 * the operation returns no result.
+	 *
+	 * @link https://github.com/UnionOfRAD/lithium/issues/1210
+	 */
+	public function testRawOptionSettingWithNoResultResource() {
+		$expected = array();
+		$result = Galleries::connection()->read('SET SESSION group_concat_max_len = 100000;');
```
davidpersson: As the query is actually executed, following tests might be affected by the change. Is there any kind of query that fulfills our test purposes but does not change the server/client behavior?

shopblocks: Im afraid that I cannot say for certain whether there is, and I dont have time to check this. Perhaps we can set a session on something that doesn't actually matter? Or perhaps reset the session afterwards?

koushki: I think this is similar `SET @temp = 1`, could you please test this one ?

shopblocks: My Tests show that this works too! Thanks @koushki 

Commit made to change this.

shopblocks:
@davidpersson Managed to get this to not change settings on the server

davidpersson:
Good. Can you squash the commits together and force push? Then we should be ready to merge. Thanks!

shopblocks:
@davidpersson Done.

alexbowers:
@davidpersson Hey, any chance of a merge? Thanks.

davidpersson:
This is currently on hold. We will merge this when release plans have been sorted out.

davidpersson:
Fixed and merged in 191d29589911fba1517e163eb72bacdf2f948e5c

alexbowers:
Thanks.

Issue #1273, opened 2016-05-07
Database connection to mysql does use unix_socket options


josh4trunks:
This below block has a few issues.
https://github.com/UnionOfRAD/lithium/blob/master/data/source/database/adapter/MySql.php#L159-163

It looks like it's trying to split host (i.e. host = localhost:/tmp/mysql.sock) while `data/Connections.php` says to pass your parameters as host, port, socket. In reality PDO expects unix_socket.

I think the best solution is to not even do the explode, and just pass host/port/unix_socket as (somewhat) documented.

davidpersson:
> I see multiple problems here, we have to resolve this in a broader scope. Connections::add should not know about host, login, password, but subclasses of Source should handle this. Users should be able to go to the API docs of their driver class and there see any required/optional settings. DSN is an internal implementation detail. unix_socket should be supported through a socket config setting. We should also deprecate usage of : notation in all Database subclasses.

davidpersson:
With the existing code does passing a socket path in `'host'` work for you?

josh4trunks:
I'll test if host:port or host:socket works. if that's how you prefer it be done I can refacture the code. that's how i've help implement it before in owncloud.

but I doubt it will works from looking at how this passes parameter to PDO
also why is 3306 even there? just dont pass a port if you dont know it

josh4trunks:
@davidpersson Do you prefer things are passed as **host:socket** / **host:port**?

Here's the logic used in owncloud, since ctype digit might not be available we can use the slightly less efficient is_numeric().
https://github.com/owncloud/core/blob/master/lib/private/DB/ConnectionFactory.php#L159-170

josh4trunks:
@davidpersson does not work with the existing code.
`'host'       => localhost:/mnt/mysql/mysql.sock,`

What was wrong with the PR I created? As far as I can understand it does implement what you were saying? Maybe you want the logic in a different file, or additional documentation?

davidpersson:
Thanks for testing and providing more information

Regarding the PR: This needs to be solved not only in MySql but in other data sources and all other classes taking host and socket. As this became more of framework wide conceptual decision I started working on it.

josh4trunks:
Ok, sounds good. Tell me if you need any additional contributions/testing for the mysql driver.

davidpersson:
Fixed in ef9eaee871caf4ec874389b559d8b8c8dda918b2

josh4trunks:
[...]

PR #1274, opened 2016-05-13
Use $port or $socket set in config

josh4trunks:
This actually uses the **$port** and **$socket** variables set in the config.
It also no longer hardcodes the port as 3306, and instead would use whatever value PDO is configured to fallback to.

### DIFF
--- a/data/source/database/adapter/MySql.php
+++ b/data/source/database/adapter/MySql.php
@@ -164,10 +164,20 @@ class MySql extends \lithium\data\source\Database {
 	 */
 	public function connect() {
 		if (!$this->_config['dsn']) {
-			$host = $this->_config['host'];
-			list($host, $port) = explode(':', $host) + array(1 => "3306");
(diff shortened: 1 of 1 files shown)


josh4trunks:
Fixes https://github.com/UnionOfRAD/lithium/issues/1273

josh4trunks:
@davidpersson would you happen to be the best person to bring this up to?

davidpersson:
I see multiple problems here, we have to resolve this in a broader scope. Connections::add should not know about host, login, password, but subclasses of Source should handle this. Users should be able to go to the API docs of their driver class and there see any required/optional settings. DSN is an internal implementation detail. unix_socket should be supported through a socket config setting. We should also deprecate usage of <host>:<port> notation in all Database subclasses.

josh4trunks:
sorry what do you mean by "unix_socket should be supported through a socket config setting"? this PR works by having a user set $socket in the config, and passes this as unix_socket to PDO.

what variable do you instead want to be set in the config?

PR #1288, opened 2016-06-30
Clarify APC deprecation message

alexanderhaensch:
Best would be to include a link to https://github.com/UnionOfRAD/framework into the error message.
Unsure about the best practice


davidpersson:
Thank you for the patch, @alexanderhaensch. I've amended your commit with a link to the framework distro in 65b00b0e9786b6be2bec17ce0b1b4efbcb11429d

Issue #1272, opened 2016-04-21
Provide autoBind option to bind filter functions to context


vbresults:
The deprecated notice on the old filter implementation says to use static:: instead of $self, since that is now gone, but the filter closures are never bound to the static class in question.

So, then, the class referred to in static:: ends up being `lithium\core\Libraries` or nothing, completely ignoring the correct static class/context.   Ping @jperras @davidpersson @nateabele @gwoo

davidpersson:
Thanks for submitting this report. Can you paste the exact error message and context the error is triggered in?

vbresults:
`PHP Fatal error:  Call to undefined method lithium\core\Libraries::read() in /var/www/com.vbresults.apps.ci/libraries/xxx/config/bootstrap/yyy.php on line ...`

```PHP
use lithium\aop\Filters;

Filters::apply('app\extensions\adapter\Zzz', 'delete', function($params, $next) {
	foreach ($params['profiles'] as $profile):
		if (!($file = static::read(compact("profile")))):
			continue;
		endif;

		unlink($file);
	endforeach;

	return $next($params);
});
```

PHP `5.6.11`

davidpersson:
app\extensions\adapter\Aaa::read() is static or concrete?

vbresults:
Right, static

davidpersson:
`Filters` will not auto-bind context of the filters (closures). This is i.e. because filters can originate from a static method and target a concrete method. This also fits well with our no automagic approach.

In your case you should use the (fully classified) class name instead of "static" - as you create the filter outside any class scope (in bootstrap phase I persume):

```
Filters::apply('app\extensions\adapter\Aaa', 'delete', function($params, $next) {
    foreach ($params['profiles'] as $profile):
        if (!($file = \app\extensions\adapter\Aaa::read(compact("profile")))):
            continue;
        endif;

        unlink($file);
    endforeach;

    return $next($params);
});
```

vbresults:
@davidpersson This isn't an adequate replacement for the removed $self variable -- at the very least, can the Filters class be configured with an option to auto-bind?  Something like `\lithium\aop\Filters::config("default", Filters::config("default") + array('auto_bind' => true));`

davidpersson:
The documentation (manual) will be updated. 
I'll look into if auto-binding is feasible, if not will document the why not.

davidpersson:
@vbresults Meanwhile this might work:
```
[...]

PR #1295, opened 2016-07-16
Document simplified filters context.

vbresults:
Ping @davidpersson 

https://github.com/UnionOfRAD/framework/pull/41#issuecomment-233122913

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,12 +48,16 @@
   special methods or properties defined. This is possible by entirely relying on a central
   filters manager and using `spl_object_hash()` internally.
 
+  **Simplified filters context.** By integrating PHP 5.5's new `::class` keyword with `use`,
+  references to the static context are now made with `Context::class`, removing the need
+  for a static `$self` and repeating the full class name across filters in the same file.
+
   **Simplified filters signature.** By using PHP 5.4's new context binding feature for
   closures, we were able to simplify the signature of filters - i.e. by dropping the
   `$self`.
 
   This - as a sideeffect - reduces the requirement of using `invokeMethod()` to access
-  protected members of the context. `$this` and `static` can be used to access
+  protected members of the context. `$this` and `Context::class` can be used to access
   the filtered object. Also makes better stacktraces.
 
   **Simplified chain advancing.**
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -48,12 +48,16 @@
...
+  references to the static context are now made with `Context::class`, removing the need
+  for a static `$self` and repeating the full class name across filters in the same file.
+
   **Simplified filters signature.** By using PHP 5.4's new context binding feature for
   closures, we were able to simplify the signature of filters - i.e. by dropping the
   `$self`.
 
   This - as a sideeffect - reduces the requirement of using `invokeMethod()` to access
-  protected members of the context. `$this` and `static` can be used to access
+  protected members of the context. `$this` and `Context::class` can be used to access
```
davidpersson: Lets keep mentioning `static` as both approaches work. Something along the lines of "... $this and static/Context::class can be used..."

davidpersson:
The following code example block will also need a change:
```
// always use this
Filters::apply('lithium\storage\Session'
```

davidpersson:
Great that you found out about using "::class" this is way shorter and sweeter.

CHANGELOG.md
vbresults: `static` does not work out of the box, it isn't automagically bound to the context class, and defaults to `lithium\core\Libraries` which in 99.9% of cases doesn't work the same;

https://github.com/UnionOfRAD/lithium/issues/1272

davidpersson: Yes, static can only be used when the filter was created inside a static method, they cannot later be bound so they can use `static``:
```
class Bar {

    protected static $_foo = 'foo';

    public static function giveFilter() {
        return function() {
            var_dump(static::$_foo);
        };
    }
}
```

`$this` will only work if the filter was created inside a concrete method OR you later bound the closure to an object.

My point here is that filters should mostly not have the need to interact directly with methods inside the class. They should be self contained. Yes there are some exceptions for these, you're right, let's not recommend `static` but `Context::class`.

Thanks for bringing this up.

davidpersson:
Will merge this, okay?

PR #1298, opened 2016-07-24
Adding nested strategy support.

jails:
When a gallery as many images, the relationship can be defined like so:

```php
class Galleries extends \lithium\data\Model {
	public $hasMany = array('Images');
	protected $_schema = array(...);
}
```
```php
class Images extends \lithium\data\Model {
	public $belongsTo = array('Galleries');
	protected $_schema = array(...);
}
```

With some classic DBMS, you can use the `'with'` option to eager load related datas out of the box:
```php
Galleries::all(['with' => ['Images']]);
```
This works because joins are done under the hood to fetch related data in a single query.

But for NoSQL databases all related data need to be fetched individually which can lead to either some complex code or a non optimised fetching strategy.

This PR introduces the `'with'` option for NoSQL database. So instead of using joins, related datas are fetched using additionnal queries. But instead of fetching related data individually, queries are done so that it'll require `the number of relations to embed + 1` queries.

For example:
```php
Galleries::all(['with' => ['Images']]);
```
will require 2 queries and:
```php
Galleries::all(['with' => ['Images.Tags']]);
```
will require 3 queries and so on.

Warning, conditions in the `'with'` option for NoSQL databases doesn't constraint the whole result. So you won't be able to perform queries like "I wan't only galleries which contain images of trees". The results will be all galleries with trees only as images.

This strategy can also be used for DBMS databases through the following syntax:
```php
Galleries::all([
	'with' => ['Images'],
	'strategy' => 'nested'
]);
```
This can be usefull if galleries and images are stored on differents hosts in different databases for example. Warning, in this cases the conditions in the `'with'` won't constraint the whole result too.

### DIFF
--- a/data/model/Relationship.php
+++ b/data/model/Relationship.php
@@ -8,7 +8,9 @@
 
 namespace lithium\data\model;
 
+use Exception;
 use Countable;
+use Traversable;
 use lithium\util\Set;
 use lithium\core\Libraries;
 use lithium\core\ConfigException;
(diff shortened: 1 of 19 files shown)


data/model/Relationship.php
```diff
@@ -312,6 +313,155 @@ protected function _strategies() {
...
+		return $indexes;
+	}
+
+	/**
+	 * Unsets the relationship attached to a collection en entities.
+	 *
+	 * @param  mixed  $collection An collection to "clean up".
+	 */
+	public function _cleanup($collection)
+	{
```
davidpersson: Minor CS: brace on new line

davidpersson:
Thanks, @jails that looks like it was missing and seems very useful. I am sure MongoDB users will appreciate this. I'll see if we can add this to 1.1 (which has many features already, I don't want to overwhelm anybody). We'll have to convert this to "new" array syntax and add the new file header (available beginning with 1.1). I can do this once it gets merged.

jails:
Thank you for the review @davidpersson, I appreciate. I'll probably leaved it opened for a while since it's a backport of a part of an ORM which still in alpha stage. Also the fact the dev branch is out of date I don't wan't to push on master too quickly.

davidpersson:
Work happens mostly in 1.0 and 1.1 branches. I'm using master just for merging in the last stable release from the version branches. I think the best choice is to base nested-strategies on 1.1 and than together we decide if we want to merge it in 1.1 or 1.2.

davidpersson:
Merged into 1.2 at 305c125386bd642fd44c552e427841dccb4a633b

PR #1305, opened 2016-12-06
Update Mongo driver

jails:
Hello,

This PR bring the mongodb driver back to li3.
The `MongoDb` class has been refactored to support all CRUD operations through the new mongodb driver (::find('count') is also supported through a custom aggregation query). I converted a couple of unit tests into integration tests to simplify testing.
The new `MongoDb` class is not compatible with old version of mongo servers (it's working with the 3.2 on travis, should work with 3.0 (didn't tested), and doesn't work with 2.x for sure).

The main drawback is that the new low level driver doesn't support GridFS out of the box anymore. So I removed all logic related to that.

Maybe we can take this PR as a starting point so we can move ahead and decide what to do for GridFS ?


davidpersson:
First of all: thank you jails for working on this. I probably wouldn't have had the time for it in the near future. Your deep knowledge of the data layer is obviously very valuable here.

Let me first answer the most important question on _When can we integrate and release this update?_  

Even if there weren't any visible breaking changes (GridFS) it's very likely that either we or the underlying new extension introduces (subtle) behavior changes, which would in turn require our users to update existing code. Especially as data storage is an especially sensitive topic we shouldn't risk stability.

The next possible integration point where we can make such changes is 2.0.

As there is demand to reach compatibility with newer MongoDB versions and mongo+PHP 7.0 soonish, we'll need to release 2.0 earlier as planned. 

The release schedule will be changed to release 2.0 at the end of 2017/early 2018 with the alpha release coming in the summer of 2017. Just because we can break compatibility for 2.0 does not mean that we have to. I'd like to keep the migration from 1.x to 2.x as smooth as possible.

bildschirmfoto 2016-12-21 um 15 10 35

1.2 will probably become the last release in the 1.x series. I'm planning to release both 1.1 and 1.2 next year with 6 months in between.

davidpersson:
I've just branched off 2.0. I've decided that we're not going to support GridFS through the framework in the future. It is a special feature of this database and hard to implement in a generic framework-y way. Implementing it will hold us back from getting full mongo support for PHP7.x.

d1rk:
@davidpersson Thanks for continuing on this. Now, when do you think there will be an alpha-version of 2 available? The current plan states 2018, but it has just been started... 😄

davidpersson:
The screenshot shows an oldish plan, it has been updated on http://li3.me/versions. 2.x will take a while, until its ready.

d1rk:
@jails Thank you for your effort on this. What is the roadmap to get this done? Is there a plan to finish this branch, as it would be a great addition to lithium. Thanks for your feedback.

jails:
Hello @d1rk, I'm not working actively on mongo anymore. Since the PHP low level driver doesn't support GridFS anymore, this PR is probably good to go. Last time I checked tests were passing so I think the best option to move ahead is to have people to test it in their prod env and fix bugs as they come.

davidpersson:
@d1rk kindly continues integration work on the new driver update. I've rebased the branch on top of 2.0, we are aiming to merge it into this release. @d1rk you might need to make the tests pass again, as well as look into a few issues (i.e. the casting of MongoId we talked about). Please post any questions on this PR, as it's bound to the new-mongo-driver branch, which you now have commit rights for.

Happy coding!

d1rk:
@jails Thanks for your response, some things changed a bit, but I am already working on it.
@davidpersson Thank you, I am glad I can be of help.

d1rk:
@davidpersson i pushed the code, that "works for me", including passing test cases. In this run, I have the same error on all builds, in the ConnectionTests, see here:

https://travis-ci.org/UnionOfRAD/lithium/jobs/366016266#L1614

Do you have an idea, what is happening here? I was unable to re-produce the same error on my machine, which makes it hard for me to work this error. If you have any directions you can point me to, I would be glad to follow those up. Thank you for your support on this.

jails:
Thank you for the fixes @d1rk.
The test was simply wrong. [Per adapter filters has been removed in 2.0.](https://github.com/UnionOfRAD/lithium/commit/b915925be7396955aa6bcf3c1de23d573baff78b) so the expectation was wrong and I simply updated it.

jails:
Btw for the  `ReadPreferenceConstants` I would probably do this instead:
```php
<?php

namespace lithium\data\source\mongo_db;

class ReadPreference extends \MongoDB\Driver\ReadPreference {
	const PRIMARY = "primary";
	const PRIMARY_PREFERRED = "primaryPreferred";
	const SECONDARY = "secondary";
	const SECONDARY_PREFERRED = "secondaryPreferred";
	const NEAREST = "nearest";
}
```
So you can use the same `ReadPreference` class everywhere and still have access to all constants (PRIMARY, RP_PRIMARY, NO_MAX_STALENESS, etc.).

d1rk:
Hi @jails thanks for jumping in. I really appreciate your fix and feedback on this. Regarding your last comment: I tried your proposed solution as my first approach to this problem. It did not work, due to the const of that, i was not able to overwrite these. But if you think, this should work i'll try again. Maybe i did something wrong.

jails:
@d1rk yeah I didn't notice `MongoDB\Driver\ReadPreference` was a `final` class at first glance, so bad...
So I don't have a strong opinion here. We can either stick with the "driver way" and using everywhere:
```php
$readPreference = new ReadPreference(ReadPreference::RP_PRIMARY);
```
Or introducing a new class but have constants splited in different classes.
It's up to you @davidpersson (with great power comes great responsibility :wink:!)

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 ### Changed
 
+- The `Mongo` data source adapter has been updated to work with the `mongodb` extension
+  instead of the old `mongo` extension. This brings back full li₃ support for MongoDB 
+  on PHP 7.x. Due to how the new extension works, we do not support GridFS through the 
+  adapter anymore. You can still access GridFS using the underlying connection object. 
+  (Simon Jaillet)
```
davidpersson: @d1rk Don't forget to add yourself here.

data/source/MongoDb.php
```diff
@@ -212,53 +213,69 @@ public static function enabled($feature = null) {
...
+				'maxPoolSize'              => null,
+				'minPoolSize'              => null,
+				'maxIdleTimeMS'            => null,
+				'waitQueueMultiple'        => null,
+				'waitQueueTimeoutMS'       => null,
+				'w'                        => WriteConcern::MAJORITY,
+				'wTimeoutMS'               => 10000,
+				'journal'                  => true,
+				'readConcernLevel'         => ReadConcern::LOCAL,
+				'readPreference'           => ReadPreferenceConstants::RP_PRIMARY,
```
davidpersson: It's unfortunate that they don't map those constants to string values, as they do with the other ones. 

I'd prefer a small conversion method that converts the read preferences constants into strings values in a generic way (`strtolower -> camelCase`). If that utility receives string values it simply returns them unmodified. The goal is to allow users to use both string values and driver constants. 

The conversion should happen around L276, where other config mangling already takes place.

data/source/MongoDb.php
```diff
@@ -212,53 +213,69 @@ public static function enabled($feature = null) {
...
-			'readPreference' => null,
-			'autoConnect' => false,
-			'dsn' => null
+			'host'               => static::DEFAULT_HOST . ':' . static::DEFAULT_PORT,
+			'login'              => null,
+			'password'           => null,
+			'database'           => null,
+			'timeout'            => 100,
+			'schema'             => null,
+			'uriOptions'            => [
```
davidpersson: I fear that the MongoDB maintainers will add more options over time, requiring us to add more options here too. Also it becomes unclear which options we overwrite. Please better remove those options -- especially `null` ones.

data/source/MongoDb.php
```diff
@@ -212,53 +213,69 @@ public static function enabled($feature = null) {
...
+				'authSource'               => null,
+				'authMechanism'            => null,
+				'gssapiServiceName'        => null,
+				'localThresholdMS'         => null,
+				'serverSelectionTimeoutMS' => null,
+				'serverSelectionTryOnce'   => null,
+				'heartbeatFrequencyMS'     => null,
+				'uuidRepresentation'       => null
+			],
+			'driverOptions' => [
```
davidpersson: Can we remove unused empty options here too?

data/source/MongoDb.php
```diff
@@ -292,11 +311,19 @@ protected function _init() {
...
 		}
-		parent::_init();
+
+		$server = $this->_classes['server'];
+
+		$notNull = function($value){
+			return $value !== null;
+		};
+		$uriOptions = array_filter($this->_config['uriOptions'], $notNull);
+		$driverOptions = array_filter($this->_config['driverOptions'], $notNull);
```
davidpersson: The above lines dealing with filtering, can be removed, once empty options are removed.

data/source/MongoDb.php
```diff
@@ -339,9 +366,6 @@ protected function _init() {
 	 * @return void
 	 */
 	public function __destruct() {
```
davidpersson: Since this method becomes entirely empty, can we remove it altogether?

data/source/MongoDb.php
```diff
@@ -363,58 +387,22 @@ public function configureClass($class) {
...
-
-			if ($this->connection = $this->server->{$this->_config['database']}) {
-				$this->_isConnected = true;
-			}
-		} catch (Exception $e) {
-			throw new NetworkException("Could not connect to the database.", 503, $e);
-		}
-		return $this->_isConnected;
+		$this->_isConnected = true;
+		return true;
```
davidpersson: Shortcut this to `return $this->_isConnected = true;` to match style in other places.

data/source/MongoDb.php
```diff
@@ -425,9 +413,12 @@ public function disconnect() {
...
 	public function sources($class = null) {
-		$this->_checkConnection();
-		$conn = $this->connection;
-		return array_map(function($col) { return $col->getName(); }, $conn->listCollections());
+		$collections = $this->server->executeCommand($this->_config['database'], new Command(['listCollections' => 1]));
+		$names = [];
+		foreach ($collections as $collection) {
+			$names[] = $collection->name;
+		}
+		return $names;
```
davidpersson: Can't array_map (which looks nicer to me) be used because `$collections` isn't an array?

data/source/MongoDb.php
```diff
@@ -212,53 +213,69 @@ public static function enabled($feature = null) {
...
 	 *        - `'database'` _string_: The name of the database to connect to. Defaults to `null`.
 	 *        - `'timeout'` _integer_: The number of milliseconds a connection attempt will wait
 	 *          before timing out and throwing an exception. Defaults to `100`.
 	 *        - `'schema'` _\Closure_: A closure or anonymous function which returns the schema
 	 *          information for a model class. See the `$_schema` property for more information.
-	 *        - `'gridPrefix'` _string_: The default prefix for MongoDB's `chunks` and `files`
-	 *          collections. Defaults to `'fs'`.
-	 *        - `'replicaSet'` _string_: See the documentation for `Mongo::__construct()`. Defaults
-	 *          to `false`.
-	 *        - `'readPreference'` _mixed_: May either be a single value such as Mongo::RP_NEAREST,
```
davidpersson: Including docs for every driver and uri option is too much work. We should point users to the general mongo documentation instead.

d1rk:
@davidpersson Thank you for your time and feedback. I remember, that we talked about the default values for uriOptions. The rest of the feedback is also valuable and I will incorporate it, soon.

davidpersson:
We talked about the uriOptions but honestly I maybe did not get of all it. Can you clear things up (again)? Are all options required?

d1rk:
@davidpersson I just learned that [MongoDB 3.4 changes how aggregation methods are to be used](https://docs.mongodb.com/v3.4/reference/command/aggregate/). Since this used for the count-calculation, we need to:

- [ ] **1**. check against MongoDB 3.6 (and 3.4 as well)
- [ ] **2**. implement the correct behavior for every version

I will now make changes to the travis-file in order to have **1** and afterwards work on **2**.

FYI: I opened a PR travis-ci/apt-source-whitelist#381 to support MongoDB 3.6 via the apt add-on of travis.

d1rk:
We now have [successfully failing tests](https://travis-ci.org/UnionOfRAD/lithium/jobs/367157428#L1634), which I can then work on to fix. I am unsure, which MongoDB version is used, but expect it to be the latest. I added 3 versions to that file, unintentionally hoping to have all of them checked in parallel. But that is obviously not the case. Since the errors are being thrown as expected, I would expect the latest version to be installed, which is 3.6 in this case.

@davidpersson How do we handle this? Do we check against 3.6 (latest) or do we need to check 3.4 and 3.2 as well? And, if so, do we need to test those on PHP7 and PHP7.1 altogether? I am unsure, if we should bloat the test-matrix. Based on mysql I can see only one version to be tested, for example. Thanks in advance for your input.

davidpersson:
> @davidpersson How do we handle this? Do we check against 3.6 (latest) or do we need to check 3.4 and 3.2 as well? And, if so, do we need to test those on PHP7 and PHP7.1 altogether? I am unsure, if we should bloat the test-matrix. Based on mysql I can see only one version to be tested, for example. Thanks in advance for your input.

In think we can/should actually test against single MongoDB version. Because of test matrix and version support bloat. Let's keep this simple. 

I suggest using 3.4 as this is currently in the distros https://packages.gentoo.org/packages/dev-db/mongodb and https://packages.ubuntu.com/artful/mongodb-server. If we can already expect breakages with version 3.4+ up until 3.6 we will use that one as our minimum requirement. 

@d1rk Do you think 3.4 will work for us?

In general I'd like to keep little dependencies on server versions as possible. That worked out nicely in the past (except for Mongo which fast moving fast :)

d1rk:
@davidpersson Yes, I think so: MongoDB 3.4 it is.

As far as I understand, MongoDB 3.4 brought some _interesting changes_ and since this version is currently in the distros, I agree we should use that version as a base-point. I change the travis config accordingly. I don't expect to many problems for close-by further versions, and if so, one could easily adapt for that, based on the experience I gathered.

We should state that MongoDB 3.4 is the minimum requirement. Where would that be?

data/source/MongoDb.php
d1rk: I understand. To be honest: I don't see any value from using the constants in the provided class over [setting a string as per documentation](https://docs.mongodb.com/manual/reference/connection-string/#read-preference-options). Users will have to lookup how to set these kind of parameters, anyway. The list of available options is: **[primary, primaryPreferred, secondary, secondaryPreferred, nearest]**.

Would you be fine, if we just pass in these config-values, without "messing around™" with them?

data/source/MongoDb.php
d1rk: I tried several versions of array_map, but since the return value of MongoDB can be varying this is actually the nicest (read: most readable) way of doing it. I say that as someone who prefers to use array_map for situations like these. 😄

davidpersson:
> We should state that MongoDB 3.4 is the minimum requirement. Where would that be?
In the docblock of the adapter is enough for now.

data/source/MongoDb.php
d1rk: @davidpersson i continue with this, without messing around with these values, too much. If there are edge-cases or problems occurring, it is easy to expand the existing code to catch some easy user-errors and/or prevent common pitfalls.

data/source/MongoDb.php
d1rk: All online documentation has been removed and references to the official php adapter and the original MongoDB documentation are in place.

data/source/MongoDb.php
```diff
@@ -30,21 +40,39 @@
...
+ *
+ * or using a DSN:
+ * ```
+ * // config/bootstrap/connections.php:
+ * Connections::add('default', [
+ *     'type'     => 'MongoDb',
+ *     'database' => 'test',
+ *     'dsn'      => 'mongodb://c0-s1.mongodb.net:27017,c0-s2.mongodb.net:27017',
+ *     //'dsn'      => 'mongodb+srv://cluster0-test.mongodb.net',
+ * ]);
```
davidpersson: First, I suggest using the same example values as the default ones we provide in https://github.com/UnionOfRAD/framework/blob/master/app/config/bootstrap/connections.php#L44
`myDb` -> `my_app`.

We shouldn't mention file locations `config/bootstrap/connections.php` in the API docs of the framework's core classes, I know one mention was in there before. Distros/apps  do not necessarily keep their configurations at the same place. These locations are standardized inside the distros only.

To show variations of configurations, I suggest to first show a full example like you do, than just pick out that single line that changes. This makes the docblock shorter and quicker to read.

```
Connections::add('default', [
    'type'     => 'MongoDb',
    'database' => 'myDb',
     'host'     => 'example.org:27017',
     'uriOptions' => ['replicaSet' => 'Cluster0'],
]);
```

Or with multiple hosts:
```
'host'   => [host1:27017, host2:27018],
```

Third, can you split up the first basic example into two fenced code blocks? That makes it clear that these are distinct cases.

data/source/MongoDb.php
```diff
@@ -30,21 +40,39 @@
...
  * By default, it will attempt to connect to a Mongo instance running on `localhost` on port
- * 27017. See `__construct()` for details on the accepted configuration settings.
+ * 27017. See the MongoDB documentation for details on the accepted configuration settings.
  *
- * This adapter is officially supported on PHP 5, where it simply needs the `mongo`
- * extension. Usage on top of PHP 7 is unofficially supported by using the new `mongodb`
- * extension in conjunction with a compatibility layer (i.e. `mongo-php-adapter`).
+ * This adapter is officially supported on PHP 7, where it simply needs the `mongodb`
+ * extension. MongoDB 3.4 and up is required, versions lower than this might work,
+ * but are not supoprted or tested.
```
davidpersson: Small typo. But otherwise great :)

data/source/MongoDb.php
```diff
@@ -212,91 +233,98 @@ public static function enabled($feature = null) {
...
+				'wTimeoutMS'               => 10000,
+				'journal'                  => true,
+				'readConcernLevel'         => 'local',
+				'readPreference'           => 'primary',
+				'readPreferenceTags'       => [],
+			],
+			'driverOptions' => [
+				'allow_invalid_hostname' => false,
+				'weak_cert_validation'   => false
+			]
```
davidpersson: I'd like we remove the spacing between keys and assignment operators. It's not super obviously called out in the CS (_Bad: (4) Either no space or too many spaces between key name and arrows._`).

data/source/MongoDb.php
```diff
@@ -212,91 +233,98 @@ public static function enabled($feature = null) {
...
+		if (!isset($config['uriOptions']['connectTimeoutMS']) && $config['timeout']) {
+			$config['uriOptions']['connectTimeoutMS'] = $config['timeout'];
+		}
+		parent::__construct($config);
 	}
 
 	/**
 	 * Initializer. Adds operator handlers which will later allow to correctly cast any
-	 * values. Constructs a DSN from configuration.
+	 * values. Constructs a DSN from configuration, if not given.
```
davidpersson: We'll get back to DSN handling over all data sources in another changeset. This is good enough now :)

data/source/MongoDb.php
```diff
@@ -212,91 +233,98 @@ public static function enabled($feature = null) {
...
-		} else {
-			$this->_config['dsn'] = sprintf(
-				'mongodb://%s',
-				implode(',', $hosts)
-			);
-		}
 		parent::_init();
 
+		switch (true) {
+
```
davidpersson: In this block a couple of blank lines sneaked in (below switch, throw and break).

data/source/MongoDb.php
```diff
@@ -212,91 +233,98 @@ public static function enabled($feature = null) {
...
+					$hosts[] = "{$host['host']}:{$host['port']}";
+				}
+				$this->_config['dsn'] = (!$this->_config['login'])
+					? sprintf('mongodb://%s', implode(',', $hosts))
+					: sprintf(
+							'mongodb://%s:%s@%s/%s',
+							$this->_config['login'],
+							$this->_config['password'],
+							implode(',', $hosts),
+							$this->_config['database']);
```
davidpersson: We don't use the ternary operator over multiple lines, instead use the good 'ol `if () {} else {}`.

data/source/MongoDb.php
```diff
@@ -212,91 +233,98 @@ public static function enabled($feature = null) {
...
+		}
+
+		$manager = $this->_classes['manager'];
+
+		$uriOptions = $this->_config['uriOptions'];
+		$driverOptions = $this->_config['driverOptions'];
+		$this->manager = new $manager($this->_config['dsn'], $uriOptions, $driverOptions);
```
davidpersson: You can inline `$uriOptions` and `$driverOptions` into L323 and make that one than spread over multiple lines if it gets too long (one argument per line, with opening and closing braces on their own lines). I prefer this style as we don't create temporary variables, the config option keys are descriptive on their own.

data/source/MongoDb.php
```diff
@@ -518,20 +497,18 @@ public function schema($query, $resource = null, $context = null) {
...
-		$this->_checkConnection();
-
 		$defaults = [
-			'w' => $this->_config['w'],
-			'wTimeoutMS' => $this->_config['wTimeoutMS'],
-			'fsync' => false
+			'ordered'    => true,
+			'w'          => $this->_config['uriOptions']['w'],
+			'wTimeoutMS' => $this->_config['uriOptions']['wTimeoutMS'],
+			'journal'    => $this->_config['uriOptions']['journal']
```
davidpersson: The extra spaces can be removed here, too.

data/source/MongoDb.php
```diff
@@ -540,56 +517,24 @@ public function create($query, array $options = []) {
...
+			try {
+				$writeConcern = new WriteConcern($options['w'], $options['wTimeoutMS'], $options['journal']);
+				$this->manager->executeBulkWrite("{$this->_config['database']}.{$source}", $insertQuery, $writeConcern);
 
-			if ($result === true || isset($result['ok']) && (boolean) $result['ok'] === true) {
 				if ($query->entity()) {
 					$query->entity()->sync($data['create']['_id']);
 				}
 				return true;
+			} catch (BulkWriteException $e) {
```
davidpersson: Specific Exceptions are generally preferred, well done.

data/source/MongoDb.php
```diff
@@ -599,16 +544,16 @@ protected function _saveFile($data) {
...
 	 */
 	public function read($query, array $options = []) {
-		$this->_checkConnection();
-
-		$defaults = ['return' => 'resource'];
+		$defaults = [
+			'readPreference'     => $this->_config['uriOptions']['readPreference'],
+			'readPreferenceTags' => $this->_config['uriOptions']['readPreferenceTags'],
+			'return'             => 'resource'
+		];
```
davidpersson: Extra spaces 🚔

data/source/mongo_db/Result.php
```diff
@@ -19,6 +19,8 @@
  */
 class Result extends \lithium\data\source\Result {
 
+	protected $_it = null;
```
davidpersson: Either give this a more descriptive name or add a short dockblock.

tests/integration/data/Base.php
```diff
@@ -32,7 +32,6 @@ public function connect($connection, $options = []) {
 
 	public function with($adapters) {
 		$type = $this->_dbConfig['adapter'] ?: $this->_dbConfig['type'];
-
```
davidpersson: Accidentally removing a blank line.

tests/integration/data/FieldsTest.php
```diff
@@ -103,6 +103,7 @@ public function testSingleField() {
 			'conditions' => [$key => $id],
 			'fields' => [$key, 'name']
 		]);
+
```
davidpersson: Accidentally adding a blank line. Files that are not functionally been changed, shouldn't appear in change sets.

data/source/MongoDb.php
d1rk: I took care of DSN Handling in https://github.com/UnionOfRAD/lithium/commit/85deea943b88392702c895ad3bad233935c36fb7. My goal was to support connection strings, displayed by MongoDB Atlas without any changes.

I wrote test cases for that and implemented accordingly. I thought about adding/expanding the HostString class (or create a new `DsnString`). If you want me to take care of that, I already have a branch for that.

tests/integration/data/FieldsTest.php
d1rk: Thanks for your valuable feedback. I actually can't explain how the blank line sneaked in. I usually double-check the changes before committing and avoid whitespace-changes.

data/source/MongoDb.php
davidpersson: That would be great to have consistent DSN support across all the data source adapters! But let's get this PR merged first. I don't want to hold it back with anything.

data/source/MongoDb.php
```diff
@@ -28,23 +39,47 @@
...
+ *     'type' => 'MongoDb',
+ *     'database' => 'my_app',
+ *     'host' => 'example.org:27017',
+ *     'uriOptions' => ['replicaSet' => 'Cluster0'],
+ * ]);
+ * ```
+ *
+ * Or with multiple hosts:
+ * ```
+ *     'host' => [host1:27017, host2:27018], //multiple hosts
```
davidpersson: `, //multiple hosts` can be dropped, it now is redundant information

.travis.yml
```diff
@@ -5,14 +5,20 @@
 # code is distributed under the terms of the BSD 3-Clause License.
 # The full license text can be found in the LICENSE.txt file.
 #
-
+sudo: false
```
davidpersson: Keep the blank line between the comment and code/config.
Opting into travis' container infra, will need to become a commit on its own.

data/source/MongoDb.php
```diff
@@ -84,7 +112,7 @@ class MongoDb extends \lithium\data\Source {
 		'schema'       => 'lithium\data\source\mongo_db\Schema',
 		'exporter'     => 'lithium\data\source\mongo_db\Exporter',
 		'relationship' => 'lithium\data\model\Relationship',
-		'server'       => 'MongoClient'
+		'manager'      => 'MongoDB\Driver\Manager'
```
davidpersson: Style cleanup: remove the extra whitespace between key and assignment operator.

data/source/MongoDb.php
```diff
@@ -153,14 +181,14 @@ class MongoDb extends \lithium\data\Source {
...
+	 *  id      => MongoDB\BSON\ObjectId
+	 *  date    => MongoDB\BSON\UTCDateTime
+	 *  regex   => MongoDB\BSON\Regex
 	 *  integer => integer
 	 *  float   => float
 	 *  boolean => boolean
-	 *  code    => MongoCode
-	 *  binary  => MongoBinData
+	 *  code    => MongoDB\BSON\Javascript
+	 *  binary  => MongoDB\BSON\Binary
```
davidpersson: Style cleanup: remove the extra whitespace between key and assignment operator.

data/source/MongoDb.php
```diff
@@ -153,14 +181,14 @@ class MongoDb extends \lithium\data\Source {
 	 *
 	 * The tyes in the schema map to database native type like this:
 	 * ```
-	 *  id      => MongoId
-	 *  date    => MongoDate
-	 *  regex   => MongoRegex
+	 *  id      => MongoDB\BSON\ObjectId
+	 *  date    => MongoDB\BSON\UTCDateTime
+	 *  regex   => MongoDB\BSON\Regex
```
davidpersson: Style cleanup: remove the extra whitespace between key and assignment operator.

data/source/MongoDb.php
```diff
@@ -153,14 +181,14 @@ class MongoDb extends \lithium\data\Source {
 	 *
 	 * The tyes in the schema map to database native type like this:
```
davidpersson: Small typo: `tyes` -> `types`

data/source/mongo_db/Schema.php
```diff
@@ -9,24 +9,25 @@
...
-		'timestamp'    => 'date',
-		'int'          => 'integer'
+		'MongoDB\BSON\ObjectId'    => 'id',
+		'MongoDB\BSON\UTCDateTime' => 'date',
+		'MongoDB\BSON\Javascript'  => 'code',
+		'MongoDB\BSON\Binary'      => 'binary',
+		'MongoDB\BSON\Regex'       => 'regex',
+		'datetime'                 => 'date',
+		'timestamp'                => 'date',
+		'int'                      => 'integer'
```
davidpersson: Style cleanup: remove the extra whitespace between key and assignment operator.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -39,17 +38,16 @@ class MongoDbTest extends \lithium\test\Unit {
...
 		'tags'              => ['type' => 'string', 'array' => true],
-		'comments'          => 'MongoId',
-		'authors'           => ['type' => 'MongoId', 'array' => true],
-		'created'           => 'MongoDate',
+		'comments'          => 'MongoDB\BSON\ObjectId',
+		'authors'           => ['type' => 'MongoDB\BSON\ObjectId', 'array' => true],
+		'created'           => 'MongoDB\BSON\UTCDateTime',
 		'modified'          => 'datetime',
 		'voters'            => ['type' => 'id', 'array' => true],
 		'rank_count'        => ['type' => 'integer', 'default' => 0],
```
davidpersson: Style cleanup: remove the extra whitespace between key and assignment operator.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
-		});
-		$this->assertTrue($db->disconnect());
-	}
+	public function testConnectNoHost() {
+		$config = $this->_testConfig;
+		$config['host'] = '';
+		$closure = function() use ($config) {
+			$db = new MongoDb($config);
+		};
+		$this->assertException('lithium\core\ConfigException', $closure);
```
davidpersson: Style: declare the closure inline `$this->assertException('lithium\core\ConfigException', function() { ... );`

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
+
+		$config = $this->_testConfig;
+		$config['host'] = null;
+		$closure = function() use ($config) {
+			$db = new MongoDb($config);
+		};
+		$this->assertException('lithium\core\ConfigException', $closure);
```
davidpersson: Declare closure inline.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
+	}
+
+	public function testConnectInvalidDsn() {
+		$config = $this->_testConfig;
+		$config['dsn'] = 'foobar://user:pass@example.org';
+		$closure = function() use ($config) {
+			$db = new MongoDb($config);
+		};
+		$this->assertException('lithium\core\ConfigException', $closure);
```
davidpersson: Declare Closure inline.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
+
+		$config = $this->_testConfig;
+		$config['dsn'] = 'foobar://user:pass@example.org/testdb';
+		$closure = function() use ($config) {
+			$db = new MongoDb($config);
+		};
+		$this->assertException('MongoDB\Driver\Exception\InvalidArgumentException', $closure);
```
davidpersson: Declare Closure inline.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
+			'cluster0-shard-00-01-foo.mongodb.net:27017',
+			'cluster0-shard-00-02-foo.mongodb.net:27017',
+		]);
 
-	public function testGoodConnectionBadDatabase() {
-		$db = new MongoDb(['database' => null, 'autoConnnect' => false]);
+		$config = $this->_testConfig;
+		$config['dsn'] = 'mongodb://cluster0-shard-00-00-foo.mongodb.net:27017,cluster0-shard-00-01-foo.mongodb.net:27017,cluster0-shard-00-02-foo.mongodb.net:27017/testdb';
+		$db = new MongoDb($config);
+		$this->assertEqual($db->_config['database'], 'testdb');
```
davidpersson: This looks a little too compact to me. Maybe separate the `assert`s from the initialization/setup phase.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
+			'cluster0-shard-00-02-foo.mongodb.net:27017',
+		]);
 
-		$this->assertException('Could not connect to the database.', function() use ($db) {
-			$db->connect();
-		});
+		$config = $this->_testConfig;
+		$config['dsn'] = 'mongodb+srv://user:pass@cluster0-foo.mongodb.net/testdb';
+		$db = new MongoDb($config);
+		$this->assertEqual($db->_config['login'], 'user');
```
davidpersson: This looks a little too compact to me. Maybe separate the `assert`s from the initialization/setup phase.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
+		$config['host'] = ['host1', 'host2:27017', 'host3:27018', ':27019'];
+		$db = new MongoDb($config);
+		$this->assertEqual($db->_config['dsn'], 'mongodb://host1:27017,host2:27017,host3:27018,localhost:27019');
+	}
+
+	public function testConnectOptions() {
+		$config = $this->_testConfig;
+		$config['dsn'] = 'mongodb://user:pass@cluster0-shard-00-00-foo.mongodb.net:27017,cluster0-shard-00-01-foo.mongodb.net:27017,cluster0-shard-00-02-foo.mongodb.net:27017/testdb?ssl=true&replicaSet=Cluster0-shard-0&authSource=admin';
+		$db = new MongoDb($config);
+		$this->assertEqual($db->_config['login'], 'user');
```
davidpersson: This looks a little too compact to me. Maybe separate the `assert`s from the initialization/setup phase.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -86,25 +83,99 @@ public function tearDown() {
...
+		$this->assertEqual($db->_config['password'], 'pass');
+		$this->assertEqual($db->_config['database'], 'testdb');
+		$this->assertEqual($db->_config['host'], 'cluster0-foo.mongodb.net:27017');
+	}
+
+	public function testConnectHosts() {
+		$config = $this->_testConfig;
+		$config['host'] = ['host1', 'host2:27017', 'host3:27018', ':27019'];
+		$db = new MongoDb($config);
+		$this->assertEqual($db->_config['dsn'], 'mongodb://host1:27017,host2:27017,host3:27018,localhost:27019');
```
davidpersson: This looks a little too compact to me. Maybe separate the `assert`s from the initialization/setup phase.

tests/cases/data/ConnectionsTest.php
```diff
@@ -76,7 +76,7 @@ public function testConnectionGetAndReset() {
 		$msg = "Cannot connect to localhost:{$this->_port}";
 		$this->skipIf(!$this->_canConnect('localhost', $this->_port), $msg);
 
-		$expected = $this->config + ['type' => 'database', 'filters' => []];
+		$expected = $this->config + ['type' => 'database'];
```
davidpersson: Please Re-Rebase the mongo branch onto the current `2.0` I've cherry picked that commit there, as it solve a problem on that branch.

tests/mocks/data/source/MockMongoManager.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/**
+ * li₃: the most RAD framework for PHP (http://li3.me)
+ *
+ * Copyright 2016, Union of RAD. All rights reserved. This source
```
davidpersson: Use `2016` should be the year this file was created.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -911,206 +710,11 @@ public function testGeoQueries() {
...
-		]);
-
-		$trace = $db->server->trace;
-		$this->assertEqual('__call', $trace[1][0]);
-		$this->assertEqual('setReadPreference', $trace[1][1][0]);
-		$this->assertEqual('__get', $trace[2][0]);
-		$this->assertEqual('test', $trace[2][1][0]);
-	}
-
-	public function testDefaultSafeOptions() {
```
davidpersson: Why did this test need to go? What have we been testing here?

tests/cases/data/source/MongoDbTest.php
```diff
@@ -911,206 +710,11 @@ public function testGeoQueries() {
...
 	public function testSchemaCallback() {
 		$schema = ['_id' => ['type' => 'id'], 'created' => ['type' => 'date']];
-		$db = new MongoDb(['autoConnect' => false, 'schema' => function() use ($schema) {
+		$db = new MongoDb(['schema' => function() use ($schema) {
 			return $schema;
 		}]);
 		$this->assertEqual($schema, $db->describe(null)->fields());
 	}
-
-	public function testSetReadPreference() {
```
davidpersson: Are we testing this elsewhere?

tests/cases/data/source/MongoDbTest.php
```diff
@@ -630,125 +537,18 @@ public function testRelationshipGenerationWithPluralNamingConvention() {
...
-	}
-
-	public function testSourcesNoConnectionException() {
-		$db = new MongoDb(['host' => null, 'autoConnect' => false]);
-		$this->assertException('Could not connect to the database.', function() use ($db) {
-			$db->sources(null);
-		});
-	}
-
-	public function testAtomicUpdate() {
```
davidpersson: I don't see where else we are testing atomic updates now. Why did this need to go?

tests/cases/data/source/MongoDbTest.php
```diff
@@ -630,125 +537,18 @@ public function testRelationshipGenerationWithPluralNamingConvention() {
...
-			]
-		];
-		$this->assertEqual($expected, $result);
-	}
-
-	/**
-	 * Tests that the MongoDB adapter will not attempt to overwrite the _id field on document
-	 * update.
-	 */
-	public function testPreserveId() {
```
davidpersson: Looks like a valuable test to me, are we checking for this case implicitly elsewhere now?

tests/cases/data/source/MongoDbTest.php
```diff
@@ -911,206 +710,11 @@ public function testGeoQueries() {
...
-
-		$query = new Query(['type' => 'read', 'source' => 'custom']);
-		$this->_db->delete($query);
-		$result = $this->_db->connection->custom->call;
-		$expected = ['justOne' => false, 'w' => 1, 'wTimeoutMS' => 10000, 'fsync' => false];
-		$this->assertEqual('remove', $result['method']);
-		$this->assertEqual($expected, $result['params'][1]);
-	}
-
-	public function testGridFsCRUDWithDefaultPrefix() {
```
davidpersson: This and the following tests are okay to be removed, as respondsTo is going away and we don't support GridFS anymore.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -442,122 +458,13 @@ public function testCreate() {
...
 		$this->assertTrue(MongoDb::enabled('arrays'));
 		$this->assertTrue(MongoDb::enabled('booleans'));
 		$this->assertTrue(MongoDb::enabled('relationships'));
 	}
 
-	public function testArbitraryMethodCalls() {
-		$this->assertInternalType('array', $this->_db->listDBs());
-	}
-
-	public function testDocumentSorting() {
```
davidpersson: Seems like an important test.

tests/cases/data/source/MongoDbTest.php
```diff
@@ -272,166 +448,6 @@ public function testConditionsWithSchema() {
 		$this->assertEqual($expected, $result);
 	}
 
-	public function testReadNoConditions() {
```
davidpersson: The crud operations are tested in the `Crud` base test, so as along as we don't check for any mongo specialities, they are good to be removed here.

tests/cases/data/source/MongoDbTest.php
jails: Some tests has been removed because they were more testing mocked code that actually real code. So this one for example has been converted as an integration test in `tests/integration/data/MongoDbTest.php`

davidpersson: Okay good that this moved to the right place. You're right, some unit tests - especially the `data` ones - seem like integration tests to me.

data/source/MongoDb.php
d1rk: As this is just a comment, I'd leave it as it is, as there is focus on humans readability. What do you think?

data/source/MongoDb.php
d1rk: @davidpersson see above.

d1rk:
Thanks for your feedback I still need to re-rebase this from 2.0. What about the unit-tests? I did not have enough time to look into that, but I could do that, soon.

davidpersson:
As I understand correctly the unit tests, have never been real unit tests as they've been touching the database, they now have been converted into integration tests.

d1rk:
@davidpersson I finished re-basing on 2.0 and squashing commits. I hope this is all ok, now and works as expected.

tests/cases/data/source/MongoDbTest.php
d1rk: done

tests/cases/data/source/MongoDbTest.php
d1rk: done.

davidpersson:
Hey everybody, thanks for your persistence and getting this into 2.0!

PR #1308, opened 2017-02-15
Fix for Unicode characters length

EleRam:
Fix 'lengthBetween' validator for counting Unicode (multibyte) string length.

Lithium is using `strlen()` function and it returns incorrect Unicode string length.
an example in Persian (Farsi) language: `strlen('سلام')` returns **8** which it is **4** characters.

using `mb_strlen()` can fix it.

[PHP.net: Multibyte strings - Introduction](http://php.net/manual/en/intro.mbstring.php)
[PHP.net: mb_strlen function](http://php.net/manual/en/function.mb-strlen.php)

### DIFF
--- a/util/Validator.php
+++ b/util/Validator.php
@@ -274,7 +274,7 @@ class Validator extends \lithium\core\StaticObject {
 				return in_array($value, $options['list'], $strict);
 			},
 			'lengthBetween' => function($value, $format, $options) {
-				$length = strlen($value);
+				$length = mb_strlen($value);
 				$options += array('min' => 1, 'max' => 255);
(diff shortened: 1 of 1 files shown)


EleRam:
Updated comment!

davidpersson:
@EleRam thanks for the well worded description of the problem and the proposed solution. You are correct: the default validation rule does not take into account multibyte characters. This is on purpose. Multibyte handling in PHP is still opt-in and so is our solution for the problem. Please see 
https://github.com/UnionOfRAD/framework/blob/master/app/config/bootstrap/g11n.php#L209
for how it is done when using our default app distribution.

PR #1309, opened 2017-03-14
Added HTTP Status Code 426 per RFC7231

ericcholis:
Added HTTP Status Code 426 per [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.5.15).

The 426 (Upgrade Required) status code indicates that the server refuses to perform the request using the current protocol but might be willing to do so after the client upgrades to a different protocol.


davidpersson:
Merged in ea6be5a6315a7aea41ddb4012a9bbad6d72bb4f4

PR #1316, opened 2017-04-26
welcome to 2017

d1rk:


davidpersson:
Thanks for the proposal @d1rk. It was indeed time to do something about it. I've decided to not update the copyright year each new year. Instead we will simply use the creation date of each file. This way we don't need to create sweeping commits just to update a single number.

https://github.com/UnionOfRAD/lithium/commit/4af3ad3b13a1c2488657212c4a5591feb80a574e

davidpersson:
For those interested: I used this little script:
https://gist.github.com/davidpersson/da5245c0e689b17b1162408d12e28279

d1rk:
Thank you for your solution and sharing the code for it. I like this much better.

PR #1318, opened 2017-07-25
Add support for MasterCard 2-series numbers.

gscherer:
These numbers start with 6 digits between 222100 and 272099, but otherwise work the same as the existing 5 series numbers.


davidpersson:
Thanks for submitting this improvement. Merged in c567bb711fa558c77d43bf0f6ecd8a7248b38bc5

PR #1321, opened 2017-11-07
Fix the code attempting to capture the PDO specific error code in Database::connect

gscherer:
Fix the code attempting to capture the PDO specific error code from the PDOException message. Currently, if the first capture group of the regex is empty, it reassigns `$code` to the entire message string, which is not what any of the logic following it is expecting. If the first capture group is not empty, `$code` should be assigned to its contents.

I am also removing the `var_dump` call, as I think that was probably debug code
left in the main line by accident.


gscherer:
Not sure if I need to base this change off of another branch instead of 1.1. It seems that a lot of tests are failing when I run the unit tests against the 1.1 branch.

gscherer:
This looks like it's already been fixed in the dev branch. Should have looked there first.

PR #1346, opened 2020-04-07
Bug fix for MySQL 8.0

alexbowers:
In MySQL 8.0 tinyint without a length defined returns no length, causing a notice to be thrown

### DIFF
--- a/data/source/database/adapter/MySql.php
+++ b/data/source/database/adapter/MySql.php
@@ -446,7 +446,7 @@ class MySql extends \lithium\data\source\Database {
 		switch (true) {
 			case in_array($column['type'], ['date', 'time', 'datetime', 'timestamp']):
 				return $column;
-			case ($column['type'] === 'tinyint' && $column['length'] == '1'):
+			case ($column['type'] === 'tinyint' && isset($column['length']) && $column['length'] == '1'):
 			case ($column['type'] === 'boolean'):
 				return ['type' => 'boolean'];
 			break;


alexbowers:
If you need this on your own projects, you can do it yourself.

Create a class in your own project, i've called mine `MySqlOverride`.

```php
<?php

namespace app\overrides;

use lithium\data\source\database\adapter\MySql;

class MySqlOverride extends MySql
{
    protected function _column($real) {
		if (is_array($real)) {
			return $real['type'] . (isset($real['length']) ? "({$real['length']})" : '');
		}

		if (!preg_match('/(?P<type>\w+)(?:\((?P<length>[\d,]+)\))?/', $real, $column)) {
			return $real;
		}
		$column = array_intersect_key($column, ['type' => null, 'length' => null]);

		if (isset($column['length']) && $column['length']) {
			$length = explode(',', $column['length']) + [null, null];
			$column['length'] = $length[0] ? (integer) $length[0] : null;
			$length[1] ? $column['precision'] = (integer) $length[1] : null;
		}

		switch (true) {
			case in_array($column['type'], ['date', 'time', 'datetime', 'timestamp']):
				return $column;
			case ($column['type'] === 'tinyint' && isset($column['length']) && $column['length'] == '1'):
			case ($column['type'] === 'boolean'):
				return ['type' => 'boolean'];
			break;
			case (strpos($column['type'], 'int') !== false):
				$column['type'] = 'integer';
			break;
			case (strpos($column['type'], 'char') !== false || $column['type'] === 'tinytext'):
				$column['type'] = 'string';
			break;
			case (strpos($column['type'], 'text') !== false):
				$column['type'] = 'text';
			break;
			case (strpos($column['type'], 'blob') !== false || $column['type'] === 'binary'):
				$column['type'] = 'binary';
			break;
			case preg_match('/float|double|decimal/', $column['type']):
				$column['type'] = 'float';
			break;
			default:
				$column['type'] = 'text';
			break;
		}
		return $column;
	}
}
```

And in `connections.php`, change your connection `adapter` to reference the class.

```php
use app\overrides\MySqlOverride;

Connections::add(
    'default', [
        'persistent' => false,
        'type'     => 'database',
        'adapter'  => MySqlOverride::class,
        'host'     => getenv('database.host'),
        'database' => getenv('database.name'),
        'login'    => getenv('database.user'),
        'password' => getenv('database.pass'),
        'encoding' => 'UTF-8'
    ]
);
```

nateabele:
This is fixed in 03719f59c2055884e29c2fd70d3c7ae50762c118. Thanks!

PR #1367, opened 2023-08-22
Code quality fixes (docblocks and function signatures)

adamroyle:


data/Collection.php
```diff
@@ -104,7 +104,7 @@ abstract class Collection extends \lithium\util\Collection {
 	 * will be assigned here.
 	 *
 	 * @see lithium\data\Schema
-	 * @var lithium\data\Schema
+	 * @var \lithium\data\Schema
```
nateabele, line 107: Our thing was always not to do leading `\` since it's (AFAIK) always implied and just looks like noise. Is there a compelling reason to change direction on that? If so, we should probably just reformat the entire codebase in one shot, rather than leave it inconsistent.

nateabele:
One comment / question—otherwise looks good.

adamroyle:
I was getting some warnings with intellisense in vscode. It needs the leading backslash otherwise it thinks it's a relative namespace to the current file. 

I searched the codebase for other "@var lithium\" and "@return lithium\" and added the leading slash. I didn't search "@param" though so there might be some of those hanging around.

nateabele:
@adamroyle That makes sense. Then yeah, throw in `@param` and we’ll merge this. Thanks man!

adamroyle:
Have pushed another commit that addresses all the remaining namespace issues I could find.

In doing these, I've now noticed a bunch of warnings for incorrect or incomplete return types. (eg. `@return array` where it should be `@return array|null`). I'm not going to worry about these unless it impacts my app. A lot of these are on `protected` functions or only for edge cases.

nateabele:
Yeah, sounds good. Thank you sir.

